PR #111, opened 2026-07-11
Add gradient projection for zerosum gauge in StandardizedRBM and CenteredRBM

### LINKED ISSUE #110: Add gradient projection `zerosum!(∂::∂RBM, rbm::StandardizedRBM)` to match the fixed standardized gauge
## Context

#108 (released in v5.5.0) made `zerosum!(::StandardizedRBM)` an exact gauge transformation, with the gauge condition on the *unstandardized* parameters (the stored standardized weights need not be zero-sum over colors, since the interaction involves `(v − offset_v) / scale_v`).

The gradient projection, however, still exists only for plain RBMs: `zerosum!(∂::∂RBM, rbm::RBM)` in `src/gauge/zerosum.jl`. There is no `StandardizedRBM` counterpart, and the two training paths now differ:

- plain `pcd!` (`src/train/pcd.jl:92`) projects the gradient every iteration via `∂regularize!(∂, rbm; …, zerosum)`;
- standardized `pcd!` (`src/standardized.jl:459`) calls `∂regularize!` without the `zerosum` flag — only the parameter reset `zerosum!(rbm)` runs (init + after each update).

## Why the projection matters (even though the reset alone is exact)

The reset keeps the model in gauge and is sufficient for *correctness* — every reset is a pure gauge transformation, so fantasy chains stay valid. What the projection adds:

1. **Regularization semantics.** The reset doesn't discard the gauge component of an update — it converts it into compensating field shifts (that's what makes it exact). Without projection, the gauge component of the L1/L2 weight-penalty gradient therefore ends up as field drift instead of vanishing: the weight penalty partly acts as an uncontrolled field update. With projection, penalties act purely within the gauge slice, i.e. on the canonical representative.
2. **Optimiser-state consistency.** The gauge component of the gradient is persistent even without regularization: the visible-field gradient is automatically zero-sum (`⟨v⟩` sums to 1 in both phases), but the weight gradient is not — its color-sum is `⟨h⟩_data − ⟨h⟩_model ≠ 0`. So Adam accumulates moments for a direction the reset removes every iteration; update and reset fight, and each reset is a parameter jump the optimiser state doesn't see. With projection, updates stay on the gauge slice to first order and resets shrink to Adam's second-order leakage.

## Request

A `zerosum!(∂::∂RBM, rbm::StandardizedRBM)` consistent with the v5.5.0 gauge convention, i.e. projecting onto the tangent of the constraint set `mean(w ./ scale_v; dims=1) = 0` (and the hidden-Potts analogue via `scale_h`), plus zero-sum of the stored Potts `θ` — with `PottsGumbel` forwarding methods mirroring the plain-RBM ones. Then the standardized `pcd!` could pass `zerosum` through to `∂regularize!` like the plain path does.

Defining property (plain-SGD check, analogous to the existing plain-RBM behavior): starting from `zerosum!(srbm)` with nontrivial offsets/scales, after

```julia
∂ = ∂free_energy(srbm, v)  # + any ∂regularize!
zerosum!(∂, srbm)          # requested method
# naive SGD step on (visible.par, hidden.par, w)
```

the model is still in gauge *without* needing the reset: `mean(srbm.w ./ srbm.scale_v; dims=1) ≈ 0` and `mean(srbm.visible.θ; dims=1) ≈ 0`.

## Downstream motivation

StackPCD.jl maintains the zero-sum gauge of a Potts bottom RBM during joint training exactly as `pcd!` does (init + per-update projection + reset). With #108 it can enable the gauge for standardized bottom RBMs, but only the reset half — the projection has no upstream method, so weight penalties on a standardized Potts bottom leak their gauge component into hidden fields, unlike the plain-RBM path.

cossio:
Closes #110

## Summary

Adds gradient projection methods `zerosum!(∂, ::StandardizedRBM)` and `zerosum!(∂, ::CenteredRBM)` that ensure gradient steps preserve the zerosum gauge of the equivalent unstandardized/uncentered RBM. This allows training with regularization and gauge constraints without needing periodic `zerosum!(rbm)` resets.

## Key Changes

- **`src/gauge/zerosum.jl`**: Fixed dimension indices in `zerosum!(∂, ::RBM)` — Potts color dimension is `dim=2` (not `dim=1`) in the layer `par` layout where `dim=1` indexes parameter types. The old code subtracted the mean over the singleton parameter-type dimension, which zeroed the Potts field gradients entirely, silently freezing Potts fields during `pcd!`/`ucd!` training whenever `zerosum=true` (the default).

- **`src/standardized.jl`**: 
  - Added `zerosum!(∂::∂RBM, rbm::StandardizedRBM)` that projects gradients to preserve the unstandardized gauge: removes the gauge direction `ξ .* scale_v` from weights and `ζ .* scale_h` from weights for Potts visible/hidden layers respectively.
  - Added `zerosum::Bool = false` keyword to `∂regularize!` for StandardizedRBM; when `true`, applies the gradient projection after regularization.
  - Updated `pcd!` to pass the `zerosum` flag to `∂regularize!`.

- **`src/centered.jl`**:
  - Added `zerosum!(∂::∂RBM, rbm::CenteredRBM)` that delegates to the plain RBM projection (since uncentering leaves weights unchanged).
  - Added `zerosum::Bool = false` keyword to `∂regularize!` for CenteredRBM.
  - Updated `pcd!` to pass the `zerosum` flag to `∂regularize!`.

- **`test/gauge/zerosum.jl`**: 
  - Corrected existing tests to use `dims=2` for Potts color dimension, and added assertions that the projected field gradients are not discarded wholesale.
  - Added comprehensive test suites for `zerosum!(∂, ::StandardizedRBM)` covering visible Potts, hidden Potts, both Potts, and PottsGumbel cases, including the defining property from #110: a naive SGD step with the projected gradient stays in gauge without the `zerosum!(rbm)` reset.
  - Added tests for `zerosum!(∂, ::CenteredRBM)`.
  - Added tests verifying that `∂regularize!` with `zerosum=true` maintains gauge after gradient steps.
  - Added tests confirming the projection reduces to plain RBM behavior for trivial offsets/scales and is a no-op for non-Potts layers.

- **`test/train.jl`**: Added test for standardized PCD training on Potts data, verifying that the gauge of the equivalent unstandardized RBM is maintained after training.

- **`test/jlarrays.jl`**: Added GPU compatibility test for `zerosum!(∂, ::StandardizedRBM)` on both-Potts models.

- **`CHANGELOG.md`**: Documented the new feature and the fix under `## Unreleased`.

- **`Project.toml`**: Bumped version to `5.6.0-DEV`.

## Implementation Details

The gradient projection for StandardizedRBM accounts for the relationship between standardized and unstandardized parameters. For Potts layers, the gauge condition on the unstandardized model (e.g., `sum(w ./ scale_v; dims=1) == 0`) translates to removing the component `mean(∂.w ./ scale_v; dims=1) .* scale_v` from the weight gradient. This ensures that a gradient step with the projected gradient preserves the gauge of the equivalent unstandardized RBM, even when combined with regularization.

https://claude.ai/code/session_01M7VVAqKypVR5kL2jJ6BypF

### DIFF
--- a/src/gauge/zerosum.jl
+++ b/src/gauge/zerosum.jl
@@ -68,11 +68,11 @@ Projects the gradient so that it doesn't modify the zerosum gauge.
 function zerosum!(∂::∂RBM, rbm::RBM)
     # ∂.visible and ∂.hidden have the layer `par` layout, where dim 1 indexes the
     # parameter type (θ, singleton for Potts) and dim 2 the Potts colors.
-    if rbm.visible isa Potts
+    if rbm.visible isa Union{Potts, PottsGumbel}
         zerosum!(∂.visible; dims = 2)
         zerosum!(∂.w; dims = 1)
     end
-    if rbm.hidden isa Potts
+    if rbm.hidden isa Union{Potts, PottsGumbel}
         zerosum!(∂.hidden; dims = 2)
         zerosum!(∂.w; dims = ndims(rbm.visible) + 1)
     end
@@ -125,10 +125,6 @@ function zerosum!(rbm::RBM{<:AbstractLayer, <:PottsGumbel})
     _rbm = zerosum!(gumbel_to_potts(rbm))
     return RBM(_rbm.visible, PottsGumbel(_rbm.hidden), _rbm.w)
 end
-
-zerosum!(∂::∂RBM, rbm::RBM{<:PottsGumbel, <:PottsGumbel}) = zerosum!(∂, gumbel_to_potts(rbm))
-zerosum!(∂::∂RBM, rbm::RBM{<:AbstractLayer, <:PottsGumbel}) = zerosum!(∂, gumbel_to_potts(rbm))
-zerosum!(∂::∂RBM, rbm::RBM{<:PottsGumbel, <:AbstractLayer}) = zerosum!(∂, gumbel_to_potts(rbm))
 
 zerosum_weights(weights::AbstractArray, rbm::RBM{<:PottsGumbel, <:PottsGumbel}) = zerosum_weights(weights, gumbel_to_potts(rbm))
 zerosum_weights(weights::AbstractArray, rbm::RBM{<:AbstractLayer, <:PottsGumbel}) = zerosum_weights(weights, gumbel_to_potts(rbm))
--- a/src/gauge/zerosum.jl
+++ b/src/gauge/zerosum.jl
@@ -66,12 +66,14 @@ end
 Projects the gradient so that it doesn't modify the zerosum gauge.
 """
 function zerosum!(∂::∂RBM, rbm::RBM)
+    # ∂.visible and ∂.hidden have the layer `par` layout, where dim 1 indexes the
+    # parameter type (θ, singleton for Potts) and dim 2 the Potts colors.
     if rbm.visible isa Potts
-        zerosum!(∂.visible; dims = 1)
+        zerosum!(∂.visible; dims = 2)
         zerosum!(∂.w; dims = 1)
     end
     if rbm.hidden isa Potts
-        zerosum!(∂.hidden; dims = 1)
+        zerosum!(∂.hidden; dims = 2)
         zerosum!(∂.w; dims = ndims(rbm.visible) + 1)
     end
     return ∂
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file. The format 
 ## Unreleased
 
 - Added gradient projections `zerosum!(∂, ::StandardizedRBM)` and `zerosum!(∂, ::CenteredRBM)`, which project the gradient so that a gradient step doesn't modify the zerosum gauge of the equivalent unstandardized (uncentered) model, consistent with the gauge introduced in [#108](https://github.com/cossio/RestrictedBoltzmannMachines.jl/pull/108). `∂regularize!` for these models gains the same `zerosum` keyword as the plain-`RBM` method, and their `pcd!` trainers now pass `zerosum` through to it, projecting the gradient every iteration like the plain path does ([#110](https://github.com/cossio/RestrictedBoltzmannMachines.jl/issues/110)).
-- Fixed `zerosum!(∂, ::RBM)`, which was discarding the Potts field gradients entirely instead of projecting them: it applied the zerosum over dim 1 of the `par`-shaped gradient arrays `(1, Q, ...)`, which is the singleton parameter-type dimension rather than the color dimension, zeroing `∂.visible`/`∂.hidden`. This silently froze Potts fields during `pcd!`/`ucd!` training whenever `zerosum=true` (the default). The projection now acts on the color dimension.
+- Fixed `zerosum!(∂, ::RBM)`, which was discarding the Potts field gradients entirely instead of projecting them: it applied the zerosum over dim 1 of the `par`-shaped gradient arrays `(1, Q, ...)`, which is the singleton parameter-type dimension rather than the color dimension, zeroing `∂.visible`/`∂.hidden`. This silently froze Potts fields during `pcd!` training whenever `zerosum=true` (the default). The projection now acts on the color dimension.
 
 ## 5.5.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file. The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+- Added gradient projections `zerosum!(∂, ::StandardizedRBM)` and `zerosum!(∂, ::CenteredRBM)`, which project the gradient so that a gradient step doesn't modify the zerosum gauge of the equivalent unstandardized (uncentered) model, consistent with the gauge introduced in [#108](https://github.com/cossio/RestrictedBoltzmannMachines.jl/pull/108). `∂regularize!` for these models gains the same `zerosum` keyword as the plain-`RBM` method, and their `pcd!` trainers now pass `zerosum` through to it, projecting the gradient every iteration like the plain path does ([#110](https://github.com/cossio/RestrictedBoltzmannMachines.jl/issues/110)).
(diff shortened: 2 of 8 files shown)


src/gauge/zerosum.jl
```diff
@@ -66,12 +66,14 @@ end
 Projects the gradient so that it doesn't modify the zerosum gauge.
 """
 function zerosum!(∂::∂RBM, rbm::RBM)
+    # ∂.visible and ∂.hidden have the layer `par` layout, where dim 1 indexes the
+    # parameter type (θ, singleton for Potts) and dim 2 the Potts colors.
     if rbm.visible isa Potts
```
cossio: Shouldn't this apply also to PottsGumbel?

src/gauge/zerosum.jl
```diff
@@ -66,12 +66,14 @@ end
...
-        zerosum!(∂.visible; dims = 1)
+        zerosum!(∂.visible; dims = 2)
         zerosum!(∂.w; dims = 1)
     end
     if rbm.hidden isa Potts
```
cossio: Shouldn't this also apply also to PottsGumbel?

cossio: We probably want `isa Union{Potts,PottsGumbel}`

src/gauge/zerosum.jl
cossio: PottsGumbel was covered indirectly, via the forwarding methods `zerosum!(∂, ::RBM{<:PottsGumbel,…}) = zerosum!(∂, gumbel_to_potts(rbm))` further down the file — same pattern as the plain `zerosum`/`zerosum!` PottsGumbel variants. But checking `isa Union{Potts,PottsGumbel}` directly is cleaner: done in d498579, and the now-redundant `∂RBM` forwarders are removed (the `par` layout is identical, and the StandardizedRBM projection already used the Union check). Covered by the existing "zerosum! ∂RBM PottsGumbel variants" testset.

---

src/gauge/zerosum.jl
cossio: Done in d498579 — switched both checks to `isa Union{Potts,PottsGumbel}` and removed the now-redundant PottsGumbel forwarding methods for `zerosum!(∂, ::RBM)`.

---

cossio (commented):
## Review (at d498579)

I verified the gauge math independently and read the surrounding code (`standardize_visible`/`standardize_hidden`, `center_hidden!`, the `par` layouts). Overall this looks **correct and well-tested**. No blocking issues found.

### Correctness of the math ✔️

- **Plain `RBM` fix (`src/gauge/zerosum.jl`)**: real bug, correctly diagnosed. `∂.visible` has the `par` layout `(1, Q, N...)`, so the old `dims = 1` subtracted the array's own mean over a singleton dimension, i.e. `A .- A == 0` — the field gradient was wiped every iteration while `∂.w` was still projected correctly. The fix (`dims = 2` for fields, unchanged dims for `w`) is right, and the new regression tests in d498579 are exactly the ones that would have caught it (the "projection is the identity on `∂d - ∂m`" test is a particularly nice invariant, since one-hot sums make the PCD field gradient automatically zero-sum over colors).
- **`StandardizedRBM` projection**: checks out. With `w̃ = w ./ (scale_v ⊗ scale_h)`, the gauge orbit direction in standardized coordinates is `∝ scale_v` per (site, hidden-unit), so subtracting `ξ .* scale_v` with `ξ = mean(∂w ./ scale_v; dims = 1)` removes exactly the gauge component and enforces `sum(∂w ./ scale_v; dims = 1) == 0`. I also verified the two cross-couplings that make this non-obvious:
  - the unstandardized fields pick up `−(w ./ scale_v) * (offset_h ./ scale_h)` (and the analogous hidden term), so field-gauge preservation needs *both* the zero-sum field gradient and the projected weight gradient — which is what the implementation provides;
  - in the both-Potts case the hidden projection preserves the visible weight condition (the subtracted `ζ .* scale_h` term has zero `scale_v`-weighted color mean because `scale_h` is constant over visible indices), so the sequential application is order-independent and idempotent. The tests assert both conditions simultaneously, good.
- **`CenteredRBM` delegation**: correct. Uncentering leaves `w` unchanged and shifts fields by `−w * offset_h` (resp. `−wᵀ * offset_v`), whose color-sums vanish when `∂w` satisfies the plain zero-sum conditions, so delegating to `zerosum!(∂, RBM(rbm))` is exactly right.
- **d498579**: replacing the three `∂`-projection forwarding methods with `Union{Potts, PottsGumbel}` checks is behavior-preserving (the wrappers only rewrapped layer types; `par` layouts are identical) and removes a method-ambiguity footgun. This resolves both open review threads.

### Minor notes (non-blocking)

1. **The per-iteration `zerosum!(rbm)` reset is still load-bearing with Adam.** The projection keeps *gradients* on the gauge slice, but adaptive optimizers rescale per-coordinate, so the parameter update leaves the gauge at second order — the reset in `pcd!` (kept, correctly) is what guarantees the gauge exactly. Consequently the new `"standardized pcd potts moment matching"` gauge assertions in `test/train.jl` would pass even without the projection (the reset alone satisfies them); the tests that actually isolate the new feature are the naive-SGD-step ones in `test/gauge/zerosum.jl`. Fine as-is, just noting which test carries which claim.
2. **CHANGELOG/commit message mention `ucd!` freezing Potts fields**, but `ucd!` is only defined for `RBM{<:Binary,<:Binary}`, so Potts layers can't reach it. Harmless, but the changelog entry could drop `ucd!` for precision.
3. **Consistency**: the parameter-space `zerosum!(rbm::RBM)` still uses `isa Potts` + PottsGumbel forwarding wrappers, while the gradient method now uses the `Union` check directly. Could be unified the same way in a follow-up (the parameter-space wrappers do real conversion work, though, so it's less trivial there).
4. **Pre-existing, not from this PR**: `∂w ./ rbm.scale_v` divides by data-estimated standard deviations; a Potts color never observed at a site gives `scale_v == 0` with the default `ϵv = 0`. That already blows up the forward pass (`standardize_v`), so the projection adds no new failure mode — just flagging it since the Potts+standardized path is newly exercised.

### Tests / CI

Test coverage is thorough: dims-corrected assertions with "gradient not wholesale-discarded" guards, idempotency, reduction to the plain-RBM case for trivial offsets/scales, no-op for non-Potts, PottsGumbel, both-Potts cross-conditions, SGD-step gauge preservation for all three model types, `∂regularize!` pass-through, JLArray/GPU compatibility, and the two regression tests. CI on d498579 was still running at review time (Julia 1 / 1.10 jobs in progress); the same suite passed on the pre-PR branch push, so I expect green.

---

PR #117, opened 2026-07-13
Fix rescale_weights!(::CenteredRBM) and expand test coverage

cossio:
## Summary

This PR fixes a bug in `rescale_weights!(::CenteredRBM)` that changed the modeled distribution when hidden units are continuous and the hidden offsets are nonzero, and expands test coverage across the suite. Aside from the fix, all changes are test-only — no new layer types are introduced (`nsReLU` already exists on master; this PR only adds test coverage for it).

## Bug fix: `rescale_weights!(::CenteredRBM)`

- **Issue**: The method delegated to the plain-`RBM` version, which rescales the hidden layer and the weights but not `offset_h`. Since the centered interaction involves `h - offset_h`, this changed the modeled distribution (not a gauge transformation), corrupting `pcd!` training of centered RBMs with continuous hidden units (`rescale=true` is the default, and `pcd!` calls it every iteration).
- **Fix**: Added `rescale_hidden!(::CenteredRBM, λ)` which divides the hidden offsets by `λ` together with the activations, mirroring the `StandardizedRBM` method; `rescale_weights!` now delegates to it. Also added the `weight_norms(::CenteredRBM)` helper.
- **Behavior note**: `rescale_weights!(::CenteredRBM)` now returns a `Bool` (whether rescaling was applied) instead of the RBM, consistent with the `RBM` and `StandardizedRBM` methods. All callers use it as a statement.
- Regression tests assert the gauge invariance exactly: for every continuous hidden layer type with nonzero offsets, the joint energy is preserved under `h -> h/λ`, free energies over all enumerated visible states shift by exactly the log-Jacobian constant `sum(log, λ)` (so `p(v)` is unchanged), conditional statistics rescale correctly, and `rescale_weights!` normalizes weight norms to 1 without changing `p(v)`.

## Test additions

- **`test/gauge/zerosum.jl`**: the in-place `zerosum!` PottsGumbel testset computed `F_before` but never asserted anything with it; it now checks free-energy invariance for all three PottsGumbel variants.
- **`test/layers.jl`**: `nsReLU` routed through the generic layer-interface property loop; exact `mode_from_inputs` values for Binary/Spin/Potts (previously inequality-only); `colors`/`sitedims`/`sitesize`; law of total variance for `total_meanvar_from_inputs` across all layer types (previously dReLU-only), with and without weights; scalar kernels (`binary_var`, `binary_std`, `spin_cfg`, `spin_rand`, `relu_cfg`, `relu_rand`, `gauss_cfg`, `drelu_cgf`); ReLU as the large-`γn` limit of dReLU (the exact `γn = Inf` comparisons are NaN, which is why they are commented out).
- **`test/gauge/rescale_hidden.jl`**: no-op branches of `rescale_activations!`/`rescale_hidden!`/`rescale_weights!` for layers without scale parameters; `rescale_hidden!` free-energy invariance for dReLU/pReLU/xReLU hidden layers (previously ReLU-only).
- **`test/shift_fields.jl`**: `PottsGumbel` and `nsReLU` added to both testsets.
- **`test/regularize.jl`**: the `zerosum` keyword of `∂regularize!` matches an explicit `zerosum!(∂, rbm)` projection.
- **`test/rbm.jl`**: exact `reconstruction_error` value on a deterministic RBM.
- **`test/hdf5.jl`**: `Float32` round-trip preserves eltype.
- **`test/metropolis.jl`**: seeded the `cold_metropolis` fixed-point test (the deterministic zero-temperature map can land on a 2-cycle for an unlucky random RBM).
- Test comments describe the invariant each test verifies rather than narrating historical bugs.

https://claude.ai/code/session_01WvAuVzxuh15wxoeKFRmQwm


src/centered.jl
```diff
@@ -332,11 +332,31 @@ with the plain `RBM` ones.
 """
 zerosum!(∂::∂RBM, rbm::CenteredRBM) = zerosum!(∂, RBM(rbm))
 
+"""
+    rescale_hidden!(rbm::CenteredRBM, λ::AbstractArray)
+
+Rescale hidden unit activities by `λ`, preserving the modeled distribution.
```
cossio: Nit: "Rescale hidden unit activities by `λ`" — activations are actually *divided* by `λ` (as the plain-`RBM` docstring states precisely). This copies the ambiguous `StandardizedRBM` wording; saying "divided by `λ`" would be clearer.

---

test/layers.jl
```diff
@@ -464,3 +465,109 @@ end
     # NaN parameters propagate to a NaN mode
     @test isnan(drelu_mode(NaN, 0.0, 1.0, 1.0))
 end
+
+using RestrictedBoltzmannMachines: colors, sitedims, sitesize,
+    binary_var, binary_std, spin_cfg, spin_rand, relu_cfg, relu_rand, gauss_cfg, drelu_cgf,
+    relu_energy, mean_abs_from_inputs
```
cossio: Nit: `relu_energy` and `mean_abs_from_inputs` are already imported at the top of this file — harmless duplicates, but they could be dropped from this list.

---

cossio (commented):
## Review

The PR contains one real source change — the `rescale_weights!(::CenteredRBM)` fix in `src/centered.jl` — plus ~280 lines of new test coverage across 9 test files. I verified the fix's math against the source and checked every new exact-value assertion against the corresponding implementation. **The fix is correct and the tests are sound.**

### The bug fix: verified correct ✓

The old method delegated to the plain-`RBM` version, which reparametrizes hidden activations to `h/λ` and multiplies `w` by `λ` — but the centered interaction is `(v − offset_v)' w (h − offset_h)`, so leaving `offset_h` untouched gives `wλ(h/λ − offset_h) ≠ w(h − offset_h)` whenever `offset_h ≠ 0`. Since `pcd!(::CenteredRBM)` calls this every iteration with `rescale=true` (the default) and `center_hidden!` makes `offset_h` nonzero, this silently corrupted training with continuous hidden units. Real, high-impact bug; the diagnosis in the PR body is accurate.

The new `rescale_hidden!(::CenteredRBM, λ)` additionally does `offset_h ./= λ`, giving `wλ(h/λ − offset_h/λ) = w(h − offset_h)` — exactly invariant, and consistent with both the plain `RBM` method (`w .*= λ`) and the `StandardizedRBM` method (`offset_h ./= λ`). The regression tests check the right invariance criterion (free-energy shift equals exactly the log-Jacobian constant `sum(log, λ)`), the `offset_h` transformation directly, unit `weight_norms` after `rescale_weights!`, and the discrete-hidden no-op path.

### Notes (non-blocking)

1. **PR title/description overstate the changes**: `nsReLU` is not added by this PR — `src/layers/nsReLU.jl` and its `rescale_activations!` no-op already exist on master. The PR only adds nsReLU to test layer lists. Consider correcting the title/body so the history isn't misleading. (The CHANGELOG entry is accurate.)
2. **Return-value change**: `rescale_weights!(::CenteredRBM)` previously returned `rbm`, now returns `Bool`. I checked all callers (`pcd!` in `src/centered.jl`, `src/train/pcd.jl`, `src/train/ucd.jl`, tests) — all use it as a statement, and the new behavior matches the `RBM`/`StandardizedRBM` methods. Might deserve a clause in the CHANGELOG entry since `rescale_weights!` is documented in README/docs.
3. Two small nits posted inline (docstring wording, duplicate test imports).

### Test quality

- Exact-value assertions all match the implementations: `gauss_cfg(θ, 2) = θ²/4 + log(π)/2`, sign-of-γ invariance via `abs(γ)`, the Binary/Spin/Potts `mode_from_inputs` formulas, the `dReLU→ReLU` limit identity `cgfs(drelu) = cgfs(relu) + log1p(exp(Γn − Γp))` (matches `drelu_cgf`'s logaddexp structure), and the `reconstruction_error` semantics.
- Flakiness looks controlled: the `relu_rand` sampling check (10⁶ samples, rtol 1%) has ~7σ margin, files seed their RNGs, and the `cold_metropolis` seed addition fixes a real nondeterminism (possible 2-cycle instead of a fixed point).
- Adding `nsReLU` to `_layers` routes it through the whole generic property-test suite — the right way to cover a delegating layer type. The law-of-total-variance sweep and the free-energy-invariance checks added to the `PottsGumbel` `zerosum!` tests convert type-only assertions into actual distribution-preservation checks.

### CI

The two Julia test jobs were still in progress at review time; since the PR is almost entirely new test code, merge should wait for them — but given every hand-checked assertion matches the source and all imports resolve, I expect them to pass.

**Verdict: LGTM once CI is green**, ideally after correcting the PR title/description (note 1).

---

test/centered.jl
```diff
@@ -428,3 +428,91 @@ end
...
+]
+    rbm0 = CenteredRBM(Binary(; θ = randn(3)), hidden, randn(3, 2), randn(3), randn(2))
+    rbm1 = deepcopy(rbm0)
+    λ = 0.5 .+ rand(2)
+    @test rescale_hidden!(rbm1, λ)
+
+    # the joint energy is preserved under the reparameterization h -> h / λ
+    v = bitrand(3, 10)
+    h = randn(2, 10)
+    @test energy(rbm1, v, h ./ λ) ≈ energy(rbm0, v, h)
```
cossio: **This assertion will fail for the ReLU case.** `relu_energy` returns `+Inf` for negative activations, and `h = randn(2, 10)` almost surely contains negatives (P(all 20 draws positive) ≈ 1e-6, and this file has no RNG seed). Both sides then have `Inf` at the same samples, but array `isapprox` computes `norm(x - y)`, and `Inf - Inf = NaN`, so the comparison is `false` even though the vectors are elementwise identical.

Suggested fix: draw `h` from the model's support, e.g.

```julia
h = sample_h_from_v(rbm0, v)
```

(needs `sample_h_from_v` added to the imports), or simply `h = abs.(randn(2, 10))`, which lies in the support of all five layer types. Note line 508 (`mean_v_from_h`) is fine as-is since it's linear in `h`; only the energy comparison hits the `Inf` branch.

---

cossio (commented):
## Review of e542f41 ("Add gauge-invariance tests for rescale_hidden!(::CenteredRBM)")

Test-only commit; I re-verified the new assertions against the source.

**One real problem** (inline): the joint-energy check `energy(rbm1, v, h ./ λ) ≈ energy(rbm0, v, h)` uses `h = randn(2, 10)`, which for the **ReLU** hidden case produces `+Inf` energies on both sides (ReLU energy is `Inf` for `h < 0`), and array `isapprox` fails through `Inf - Inf = NaN`. The ReLU parametrization of this testset should fail in CI — the run currently in progress will likely confirm.

The rest checks out:
- The energy-preservation identity is exact for the other four layer types (rescaled layer energy at `h/λ` equals original at `h`; interaction `wλ · (h/λ − offset_h/λ) = w(h − offset_h)`).
- The exact-enumeration check (`collect_states` over all 8 visible states, `softmax(-F1) ≈ softmax(-F0)`) is the strongest possible invariance test for `p(v)` — nice.
- Conditional statistics scaling (`mean_h_from_v ./ λ`, `var ./ λ²`) and `mean_v_from_h` preservation are correct consequences of the reparametrization.
- All new names resolve (`collect_states` from `src/partition.jl`, `softmax` from LogExpFunctions, `Statistics.mean` already imported).
- The comment rewrites in `test/train.jl`, `test/gauge/zerosum.jl`, and `test/layers.jl` (property descriptions instead of historical bug narration) are strictly clarifying.

---

test/centered.jl
cossio: **Retracting this — I was wrong.** Julia's array `isapprox` falls back to a *component-wise* comparison when `norm(x - y)` is non-finite (see `base/floatfuncs.jl`), and scalar `isapprox(Inf, Inf)` is `true` via `==`. So matching `Inf`s pass, and CI confirms: both Julia 1.10 jobs are green on `277c03f`, which still has `h = randn(2, 10)`.

One residual (optional) point: for the ReLU case, each sample with any negative `h` entry compares `Inf == Inf`, which doesn't exercise the finite energy identity — on average only ~2.5 of the 10 samples are fully positive and compared numerically (and with ~5% probability, none are). Drawing `h` in-support (e.g. `h = sample_h_from_v(rbm0, v)`) would make every sample informative, but the identity is also covered indirectly by the free-energy checks below, so this is not blocking.

---

src/centered.jl
cossio: Fixed in 277c03f — the docstring now says activations are divided by `λ`, matching the plain-`RBM` wording.

---

test/layers.jl
cossio: Dropped the duplicate imports in 277c03f.

---

test/centered.jl
cossio: Applied the residual suggestion in d3f5ede: `h` is now drawn with `sample_h_from_v(rbm0, v)`, so it lies in the hidden layer's support and every sample exercises the finite-energy identity (no more trivial `Inf == Inf` comparisons in the ReLU case).

---

cossio (commented):
## Review of d4a9a3b + d3f5ede

Both verified against the source; no issues found.

- **d4a9a3b** (`grad2var` for `PottsGumbel`): correct. `PottsGumbel` delegates `∂cgfs`, `mean_from_inputs`, and `var_from_inputs` to `Potts` (`src/layers/pottsgumbel.jl`), so its units are one-hot with softmax means and the per-color Bernoulli variance `p(1−p)` applies exactly as for `Potts`. The test cross-checks the new method against both `var_from_inputs` and the `Potts(layer)` conversion, which pins the formula from two independent directions. The CHANGELOG note explaining why `nsReLU` gets no `grad2var` (no scale parameter → second moment not recoverable from first-order cgf gradients) matches the code structure.
- **d3f5ede** (in-support `h`): matches the suggestion, import added, comment explains the rationale. With `h = sample_h_from_v(rbm0, v)` every sample now exercises the finite-energy identity for all five layer types.

All my review threads are now resolved. Once CI is green on this head, LGTM — the only remaining (optional) item is the PR title/description still saying the nsReLU layer is added here when it only gains test coverage.

---

PR #118, opened 2026-07-13
Update tnmeanvar reference in truncnorm.jl tests

cossio:


cossio (commented):
## Review

### Overview
Tiny test-only cleanup in `test/truncnorm.jl`: adds `tnmeanvar` to the explicit `using RestrictedBoltzmannMachines: ...` import list, switches the one qualified call `RBMs.tnmeanvar(a)` to the unqualified `tnmeanvar(a)`, and drops the now-unused `import RestrictedBoltzmannMachines as RBMs` alias. Net −1 line, no behavior change.

### Correctness
- ✅ `tnmeanvar` is defined at the top level of the package module (`src/util/truncated_normal.jl`, included from `src/RestrictedBoltzmannMachines.jl`), so `using RestrictedBoltzmannMachines: tnmeanvar` resolves even though the name is not exported.
- ✅ `RBMs.` was used only at the single call site being changed (line 68), so removing the alias leaves no dangling references in this file.
- ✅ `@inferred tnmeanvar(a)` is semantically identical to `@inferred RBMs.tnmeanvar(a)` — test coverage and inference checking are unchanged.

### Style / conventions
- 👍 Consistent with the file's existing pattern of explicit narrow imports (`using RestrictedBoltzmannMachines: tnmean, tnvar, ...`), and with the test suite's standalone-module organization — each test file remains runnable on its own with `--project=test`.

### Risks
None. Test-only refactor with identical semantics; no `src/` changes, no dependency or API impact. CI (Julia 1 / 1.10 test jobs) was still running at review time but there is nothing in this diff that could plausibly fail it.

**Verdict: LGTM.**

---

Issue #124, opened 2026-07-14
Julia registration


cossio:
Permanent issue for `@JuliaRegistrator` invocations. Registrator is not triggered by PR comments, so all release registrations are posted as comments here (with `branch=release-X.Y.Z` pointing at the release commit, plus release notes). This issue stays closed on purpose — comments on closed issues still trigger Registrator. See the "Releasing a new version" section of CLAUDE.md.

cossio:
@JuliaRegistrator register branch=master

Release notes:

## Fixed

- Fixed `xReLU(layer::nsReLU)`, which built the fixed `γ = 1` array with `ones(size(layer.θ))`, always returning a host `Array{Float64}`. Since this conversion backs `energies`, `cgfs`, `∂cgfs`, `sample_from_inputs` and the other moment functions of `nsReLU`, it promoted the entire hidden-layer computation to `Float64` and, on GPU, forced a host allocation plus a host→device transfer on every call. It now uses `one.(layer.θ)`, preserving both eltype and device ([#119](https://github.com/cossio/RestrictedBoltzmannMachines.jl/issues/119)).
- Fixed `rescale_weights!(::CenteredRBM)`, which changed the modeled distribution (not a gauge transformation) when the hidden units are continuous and the hidden offsets are nonzero: it delegated to the plain-`RBM` method, rescaling the hidden layer and the weights but not `offset_h`, even though the interaction involves `h - offset_h`. This corrupted `pcd!` training of centered RBMs with continuous hidden units (`rescale=true` is the default). Added `rescale_hidden!(::CenteredRBM, λ)` and `weight_norms(::CenteredRBM)`, mirroring the `StandardizedRBM` methods. `rescale_weights!(::CenteredRBM)` now returns a `Bool` (whether the rescaling was applied) instead of the RBM, consistent with the `RBM` and `StandardizedRBM` methods.

## Added

- Added a `grad2var` method for `PottsGumbel` (same formula as `Potts`: the units are one-hot, so the variance follows from the mean). No `grad2var` exists for `nsReLU`: without a scale parameter, the second moment cannot be recovered from the first-order cgf gradients.
[...]

PR #123, opened 2026-07-14
v5.7.0

cossio:
Release version 5.7.0 with bug fixes for `nsReLU` and `CenteredRBM`.

## Summary

This release fixes two important bugs affecting GPU compatibility and correctness of centered RBM training:

1. **`xReLU(layer::nsReLU)` GPU/dtype preservation** — The fixed `γ = 1` array was constructed with `ones(size(layer.θ))`, which always returned a host `Array{Float64}`. This forced promotion of hidden-layer computations to `Float64` and caused unnecessary host allocations and transfers on GPU for `energies`, `cgfs`, `∂cgfs`, `sample_from_inputs`, and other moment functions. Now uses `one.(layer.θ)` to preserve both element type and device placement.

2. **`rescale_weights!(::CenteredRBM)` correctness** — The method was delegating to the plain-`RBM` implementation, which rescaled the hidden layer and weights but not `offset_h`. Since the interaction involves `h - offset_h`, this changed the modeled distribution (not a gauge transformation) and corrupted `pcd!` training of centered RBMs with continuous hidden units. Added `rescale_hidden!(::CenteredRBM, λ)` and `weight_norms(::CenteredRBM)` methods mirroring `StandardizedRBM`, and changed `rescale_weights!(::CenteredRBM)` to return a `Bool` (whether rescaling was applied) for consistency with other RBM variants.

## Changes

- Updated CHANGELOG.md: moved entries from `## Unreleased` to `## 5.7.0`
- Updated Project.toml: bumped version from `5.6.1-DEV` to `5.7.0`

https://claude.ai/code/session_01Y45mgEw1sVskqGUozXCc9V

PR #125, opened 2026-07-14
Document the release/registration workflow as a local register-new-version skill

cossio:
Captures the registration flow worked out during the v5.7.0 release (#123, #124) and packages it as a repo-local Claude Code skill so CLAUDE.md stays slim.

- **`.claude/skills/register-new-version/SKILL.md`** (new): the canonical, full release procedure — suggesting a semver-correct version number from the changes since the last release (the user's decision is final, with one push-back on violations); release commit; frozen `release-X.Y.Z` branch pushed at the release commit so the registration target stays pinned while master moves (Registrator issue comments only accept `branch=<name>`, not a SHA, and PR comments don't trigger it at all); the register invocation with release notes from the single permanent closed registration issue #124 (the literal trigger phrase is deliberately not written here — the bot matches it anywhere, even in backticks, as this PR's own description once demonstrated); monitoring the resulting General registry PR until AutoMerge merges it, with the restart flow (fix on master, recreate the branch, re-trigger) if AutoMerge fails or a registry maintainer requests changes; TagBot tagging; and the post-release `-DEV` bump. Commit comments remain documented as a manual fallback for a human with `gh`. Includes v5.7.0 as the worked example. (Named `register-new-version` rather than `release` to avoid conflicts with more generic skill names.)
- **CLAUDE.md**: the "Releasing a new version" section is reduced to a short pointer to the skill.
- **`docs/src/developer/testing.md`**: the human-facing copy is synchronized with the new flow (it still described the old commit-comment invocation, as flagged by review) and references the skill as canonical; the branch-deletion note sits after the monitoring step that may need to recreate the branch.

https://claude.ai/code/session_01Y45mgEw1sVskqGUozXCc9V


.claude/skills/register-new-version/SKILL.md
```diff
@@ -0,0 +1,38 @@
...
+   @JuliaRegistrator register branch=release-X.Y.Z
+
+   Release notes:
+
+   ## Breaking changes
+
+   - blah
+   ```
+
+   Caveats learned the hard way: Registrator ignores PR comments entirely ("disabled", it replies); and from an issue comment only `branch=<name>` is accepted — there is no way to pin a SHA, which is why step 2 exists. Registrator resolves the branch to a concrete SHA when it processes the comment and replies with a link to the PR it opens in the General registry. The notes are added to that registry PR and to the GitHub release. (If the register comment was posted without notes, re-invoke Registrator with them to update the registration — see [Registrator's reply](https://github.com/cossio/RestrictedBoltzmannMachines.jl/commit/a4dcb8cee859c752881c6c1bb6051edaffcecf84#commitcomment-188488609).)
```
cossio: Worth adding one more caveat here: any comment containing the literal trigger phrase fires the bot, regardless of markdown formatting — this very PR got a "disabled" reply from JuliaRegistrator just because its *description* quotes the phrase in backticks. Since the workflow routes everything through issue #124, a future comment there that merely *discusses* the process (quoting `@JuliaRegistrator register branch=...`) would trigger a real registration attempt. Suggest a sentence like: "Never quote the literal trigger phrase in comments on #124 (or anywhere) unless you intend to register — the bot matches the raw text even inside code spans."

---

.claude/skills/register-new-version/SKILL.md
```diff
@@ -0,0 +1,50 @@
...
+
+# Releasing a new version
+
+During development the version in Project.toml carries a `-DEV` suffix (e.g. `5.4.0-DEV`), and changes accumulate under an `## Unreleased` section in CHANGELOG.md.
+
+## Choosing the version number (semver)
+
+Before anything else, determine the right version number and confirm it with the user:
+
+1. Review the actual changes since the last registered version — the `## Unreleased` CHANGELOG entries **and** the commit history/diff since the previous release tag (`git log vLAST..master`), since the CHANGELOG may be incomplete. Classify them:
```
cossio, line 14: The previously documented flow — commenting `@JuliaRegistrator register` directly on the release *commit* — pins the exact SHA with no frozen branch needed, and presumably still works (v5.3.2 used it). The skill explains why PR comments and bare issue comments don't suffice, but not why the commit-comment route was dropped. If the reason is that the GitHub tooling available in Claude sessions can't post commit comments (the old CLAUDE.md used `gh api`, which remote sessions lack), it's worth stating that in one line — otherwise a future reader may "simplify" back to commit comments, or wonder whether the branch dance is cargo cult. Alternatively, mention the commit-comment route as a valid manual fallback for a human with `gh`.

---

docs/src/developer/testing.md
```diff
@@ -29,14 +29,34 @@ During development the version in `Project.toml` carries a `-DEV` suffix
...
+3. Comment on the permanent (closed) registration issue
+   [#124](https://github.com/cossio/RestrictedBoltzmannMachines.jl/issues/124):
+   `@JuliaRegistrator register branch=release-X.Y.Z`, including release notes
+   (markdown, typically the CHANGELOG entries for this version) in the same
+   comment. Registrator ignores PR comments, and issue comments accept only a
+   `branch=` target (no SHA), hence the pinned branch. The notes are added to
+   the registry PR and to the GitHub release that TagBot creates. Registrator
+   opens a PR in the General registry; once it merges, TagBot creates the
+   `vX.Y.Z` tag at the registered commit and the GitHub release automatically.
+   Afterwards the `release-X.Y.Z` branch may be deleted.
```
cossio: Ordering nit: the "branch may be deleted" note lands in step 3, *before* step 4's monitoring/restart flow — which may need to delete and recreate that same branch. The skill gets this right (deletion is its step 5, after the registry PR merges). Suggest moving this sentence to the end of step 4, e.g. "Once the registry PR has merged and TagBot has tagged, the `release-X.Y.Z` branch may be deleted." That also keeps the two documents structurally parallel, which matters since they must be kept in sync by hand.

---

cossio (commented):
## Overview

Docs-only PR (no runtime code touched): moves the release/registration procedure out of CLAUDE.md into a repo-local skill (`.claude/skills/register-new-version/SKILL.md`), leaves a short pointer in CLAUDE.md, and syncs the human-facing copy in `docs/src/developer/testing.md` with the new issue-based Registrator flow worked out during the v5.7.0 release.

## What's good

- The skill captures genuinely non-obvious, hard-won operational knowledge (Registrator ignores PR comments; issue comments accept only `branch=`, hence the frozen `release-X.Y.Z` branch; the permanent closed issue #124 trick) with a worked example and restart flow for AutoMerge failures. This is exactly the kind of thing that gets re-discovered painfully if undocumented.
- Verified at head (b0f3f77): skill name/path are consistent across all three files after the `release` → `register-new-version` rename, the frontmatter is well-formed, and the description has good trigger phrasing ("release, register, tag, or publish").
- The rename in the last commit is a good call — `release` would be prone to collision with global/plugin skills.
- CLAUDE.md stays slim, per its purpose; the skill loads on demand.

## Issues / suggestions (all minor, inline)

1. **Trigger-phrase footgun** (SKILL.md step 3): the bot matches the raw text anywhere — this PR itself got a JuliaRegistrator "disabled" reply because the description quotes the phrase. A discussion comment on #124 quoting it would fire a *real* registration. Worth a one-line warning in the skill.
2. **Missing rationale for dropping commit comments** (SKILL.md step 2): the old flow (`@JuliaRegistrator register` on the release commit) pins a SHA directly with no frozen branch; the skill doesn't say why it was abandoned, inviting future "simplification" back to it.
3. **Ordering nit in testing.md**: branch-deletion note appears in step 3, before the monitoring step that may need to recreate the branch; the skill orders this correctly.

Also: the PR description still references `.claude/skills/release/SKILL.md` — stale after the rename; consider updating it so the merged history reads cleanly.

## Risks

- Three-way manual sync (skill ↔ testing.md, plus the CLAUDE.md pointer) is an accepted maintenance burden — both files at least declare the sync obligation explicitly, which is reasonable given docs readers won't look in `.claude/`.
- Hard-coded issue/PR numbers and permalinks (#123, #124, the v5.7.0 registry PR) are fine as provenance; #124 as the operative registration target is the one load-bearing reference, and it's clearly marked permanent.
- CI: docs-only change; CodeQL passed, test jobs were still running at review time but cannot be affected by this diff.

---

cossio (commented):
## Follow-up review (b0f3f77 → 5e6e25b)

All three earlier review comments are addressed — verified against the files and marked resolved:

- **Trigger-phrase caveat**: step 3 now warns the bot matches raw text even inside backticks, with the "never quote unless registering" rule. The PR description was also rewritten to avoid the literal phrase (and the stale `.claude/skills/release/` path is fixed).
- **Commit-comment rationale**: step 2 documents the route as a manual `gh` fallback with the v5.3.2 precedent, and states the actual reason it was dropped (cloud sandbox can't post commit comments; issue comments work from any environment). Exactly the missing context.
- **testing.md ordering**: the branch-deletion note now follows the monitoring/restart step, matching the skill's structure.

New content since the last review — the **"Choosing the version number (semver)"** section — is a good addition:

- Checking `git log vLAST..master` rather than trusting the CHANGELOG, and treating the `-DEV` version as a hint rather than authoritative, are both the right calls (the v5.7.0-from-5.6.1-DEV example proves the point).
- The interaction protocol (suggest with justification → user confirms; push back once on semver violations → user's call is final) is well scoped for a skill that runs semi-autonomously.
- The breaking/feature/fix classification matches Julia registry AutoMerge expectations.

No new issues. Docs-only diff, CodeQL green. LGTM.

---

.claude/skills/register-new-version/SKILL.md
cossio: Agreed — done in 6b935d0. The caveats in step 3 now warn that the bot matches the raw trigger text anywhere in a comment or PR/issue body, even inside backticks (citing this PR's own "disabled" reply as evidence), and to never write the phrase in comments unless a registration is intended. The PR description was also rewritten to no longer contain the literal phrase.

---

.claude/skills/register-new-version/SKILL.md
cossio: Agreed — done in 6b935d0, refined in 5e6e25b. Step 2 now takes both of your suggested remedies: it states why commit comments were dropped (the sandbox that Claude cloud sessions run in cannot post them, while issue comments work generically from any environment — cloud session, local machine, or a human in the browser), and keeps the commit-comment route documented as a valid manual fallback for a human with `gh`, including the `gh api .../commits/<sha>/comments` invocation.

---

docs/src/developer/testing.md
cossio: Agreed — done in 6b935d0, with your suggested wording: the deletion note moved out of step 3 to the end of step 4, "Once the registry PR has merged and TagBot has tagged, the `release-X.Y.Z` branch may be deleted." The two documents are now structurally parallel.

---

CLAUDE.md
```diff
@@ -68,11 +68,12 @@ GPU compatibility is tested without GPU hardware in `test/jlarrays.jl`, using JL
...
+1. Make a single commit titled `vX.Y.Z` that drops the `-DEV` suffix from `version` in Project.toml and renames the `## Unreleased` CHANGELOG section to `## X.Y.Z`. Land it on `master` (directly, or via a PR like [#123](https://github.com/cossio/RestrictedBoltzmannMachines.jl/pull/123)).
+2. Create a frozen branch `release-X.Y.Z` pointing at the master commit that contains the release (the merge commit if it landed via PR) and push it: `git push origin <sha>:refs/heads/release-X.Y.Z`. Never commit to this branch — its whole point is that its HEAD *is* the release commit, so registration targets the exact commit even if master keeps moving in the meantime.
+3. Comment on issue [#124 "Julia registration"](https://github.com/cossio/RestrictedBoltzmannMachines.jl/issues/124) — the single permanent issue used for all registrations, kept closed on purpose so it doesn't pollute the issue list (comments on closed issues still trigger Registrator; do NOT open a new issue per release). Include release notes (markdown, typically the CHANGELOG entries for this version) in the same comment:
```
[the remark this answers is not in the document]
cossio: Valid finding — fixed in 5d40000, and both suggested remedies were applied. `docs/src/developer/testing.md` no longer describes the old commit-comment flow: its "Releasing a new version" section now mirrors the new procedure step-for-step (release commit, frozen `release-X.Y.Z` branch, Registrator triggered from the permanent registration issue #124, monitoring the General registry PR with the restart path, TagBot tagging, next `-DEV` bump). It also states explicitly which copy is canonical: after the later refactor in 2c76151, that's the `register-new-version` skill at `.claude/skills/register-new-version/SKILL.md`, and the docs page ends with a "keep the two in sync" note. The commit-comment route survives only in the skill, explicitly labelled as a manual fallback for a human with `gh`.

---

PR #126, opened 2026-07-14
Add CI for agent docs and Claude-powered PR review

cossio:
Adds CI to keep agent-instruction files (CLAUDE.md, AGENTS.md, and SKILL.md files under .agents/ and .claude/) honest, slim, and up-to-date with the repository, plus Claude-powered review automation for all PRs.

## Changes

- **`.github/scripts/lint_agent_docs.py`** — New deterministic linter that enforces:
  - Size budgets (CLAUDE.md/AGENTS.md ≤250 lines/16KB; SKILL.md ≤500 lines; references ≤1000 lines)
  - SKILL.md frontmatter rules per the Agent Skills spec (name format, description length, required keys)
  - All repo paths mentioned in docs actually exist (staleness check); `file.jl:42`-style line citations are supported
  - `--project=` flags point to real Julia projects with Project.toml
  - Detection of long lines duplicated verbatim across files (redundancy signal)
  - Errors fail CI, warnings are annotated

- **`.github/workflows/agent-docs.yml`** — Runs the linter on every push to master and every PR (blocking).

- **`.github/workflows/claude-pr-review.yml`** — Claude review of every non-draft PR (model `claude-sonnet-5`): reviews the full diff for bugs, correctness, math/shape issues, and CI security, posting findings as inline comments, then submits a review verdict — **approve** with justification when convinced, **request changes** only for confirmed consequential problems, comment summary otherwise. On re-runs a fresh verdict supersedes the previous one. When the diff touches agent docs, the same run also reviews them semantically (contradictions with the code, redundancy, context bloat, skill description quality). Each run reconciles threads from previous runs, resolving those addressed by new commits. The job itself never fails CI; the verdict lives in the PR review state. Runs on every push to a PR, with a concurrency group cancelling superseded runs.

- **`.github/workflows/claude.yml`** — Interactive `@claude` tag mode in issues, PR comments, reviews, and review-thread replies; gated to mentions from the owner, members, and collaborators, with the repo merge policy appended to its system prompt.

- **CLAUDE.md** — New "Pull Requests" section documenting the workflow policy: PR-authoring agents should treat the Claude reviewer's approval as the default pre-merge bar (push fixes or argue back in review threads), and must never merge or enable auto-merge — merging is reserved for the repo owner (GitHub UI or explicit instruction).

- **Minor doc updates** — Fixed stale references in `.agents/skills/restricted-boltzmann-machines-jl/references/architecture.md` and `AGENTS.md` to reflect current file organization (the pre-fix `AGENTS.md` reference to `.claude/CLAUDE.md` would have failed this very linter).

## Implementation Notes

- The linter uses only the Python 3 standard library (PyYAML is used when available for stricter YAML parsing; otherwise a minimal frontmatter parser is used); placeholder detection (`<sha>`, `$VAR`, `path/to/...`, `ext/YourExt.jl`) prevents false positives on template text
- The Claude workflows authenticate with the `CLAUDE_CODE_OAUTH_TOKEN` repo secret (subscription token from `claude setup-token`) and skip cleanly when it is unavailable (e.g. fork PRs); the action is pinned to an immutable commit SHA
- The Claude workflows cannot execute on this PR itself: `claude-code-action` self-skips on PRs whose workflow files differ from the default branch (anti-tampering). They start executing on the first PR after this one merges

https://claude.ai/code/session_01KgRbFK1rTR11VtmuCMbYTi

### DIFF
--- a/.agents/skills/restricted-boltzmann-machines-jl/references/architecture.md
+++ b/.agents/skills/restricted-boltzmann-machines-jl/references/architecture.md
@@ -21,9 +21,8 @@ Read this file when you need the package map or repo-specific edit guidance.
 - `src/rbm.jl` and `src/rbms/`
   Core RBM types plus convenience constructors and variants.
 - `src/train/`
-  Initialization, PCD, gradients, and minibatch-related training helpers.
-  Note that `src/train/minibatches.jl` exists but is not currently included by
-  the root module.
+  Initialization, PCD, gradients, and minibatch iteration helpers
+  (`src/train/infinite_minibatches.jl`).
 - `src/gauge/`
   Gauge transforms such as zero-sum, hidden rescaling, and field shifting.
 - `src/pseudolikelihood.jl`, `src/partition.jl`, `src/ais.jl`,
--- a/.agents/skills/restricted-boltzmann-machines-jl/references/architecture.md
+++ b/.agents/skills/restricted-boltzmann-machines-jl/references/architecture.md
@@ -21,7 +21,7 @@ Read this file when you need the package map or repo-specific edit guidance.
 - `src/rbm.jl` and `src/rbms/`
   Core RBM types plus convenience constructors and variants.
 - `src/train/`
-  Initialization, PCD, gradients, and minibatch iteration helpers
+  Initialization, PCD, UCD, gradients, and minibatch iteration helpers
   (`src/train/infinite_minibatches.jl`).
 - `src/gauge/`
   Gauge transforms such as zero-sum, hidden rescaling, and field shifting.
--- a/.github/scripts/lint_agent_docs.py
+++ b/.github/scripts/lint_agent_docs.py
@@ -1,0 +1,284 @@
+#!/usr/bin/env python3
+"""Lint agent-instruction files (CLAUDE.md, AGENTS.md, agent skills).
+
+Deterministic checks that keep these files honest and slim:
+
+- size budgets, so the context these files consume stays small
+- SKILL.md frontmatter rules from the Agent Skills spec
+  (https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices)
+- repo paths mentioned in the docs must exist (staleness check)
+- ``--project=<dir>`` flags must point at real Julia projects
+- long lines duplicated verbatim across files are reported (redundancy signal)
+
+Semantic checks (contradictions, redundancy in meaning, guideline adherence)
+are handled by the Claude review job in .github/workflows/agent-docs.yml.
+
+Errors fail CI (exit 1); warnings are annotated but do not fail.
+Runs with the Python 3 standard library only (PyYAML is used when available).
+"""
+
+import re
+import sys
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parents[2]
+
+LIMITS = {
+    # CLAUDE.md / AGENTS.md are loaded into context on every session.
+    "memory_lines": 250,
+    "memory_bytes": 16_000,
+    # Agent Skills spec: keep SKILL.md under 500 lines, push detail to references.
+    "skill_lines": 500,
+    "skill_name_len": 64,
+    "skill_desc_len": 1024,
+    "reference_lines": 1000,
+    # Below this length a description cannot say what the skill does *and* when to use it.
+    "skill_desc_min": 40,
+    # Lines shorter than this are too generic for the duplicate check.
+    "dup_min_len": 45,
+}
+
+MEMORY_FILES = ["CLAUDE.md", "AGENTS.md", ".claude/CLAUDE.md"]
+SKILL_GLOBS = [".agents/skills/*/SKILL.md", ".claude/skills/*/SKILL.md"]
+
+# Tokens starting with these are treated as repo paths that must exist.
+DIR_PREFIXES = (
+    "src/", "test/", "docs/", "ext/", "repl/", "notebooks/", "example/",
+    "references/", "agents/", ".claude/", ".agents/", ".github/",
+)
+# Bare filenames treated as repo-root paths (committed files only —
+# Manifest.toml is intentionally absent because it is gitignored).
+TOP_LEVEL_FILES = {
+    "CLAUDE.md", "AGENTS.md", "README.md", "CHANGELOG.md", "Project.toml",
+    "LICENSE.md", "CITATION.cff", "codecov.yml", "SKILL.md",
+}
+# Tokens containing these (case-insensitive) are placeholders, not real paths.
+PLACEHOLDER_MARKERS = ("<", ">", "$", "path/to", "your")
+
+SKILL_NAME_RE = re.compile(r"^[a-z0-9]+(-[a-z0-9]+)*$")
+KNOWN_SKILL_KEYS = {"name", "description", "version", "license", "allowed-tools", "metadata"}
+
+TOKEN_SPLIT = re.compile(r"""[\s'"()\[\]{},;`|]+""")
+PROJECT_FLAG_RE = re.compile(r"--project=([^\s'\"`,)\]]+)")
+
+errors: list[tuple[Path, int, str]] = []
+warnings: list[tuple[Path, int, str]] = []
+
+
+def error(path: Path, line: int, msg: str) -> None:
+    errors.append((path, line, msg))
+
+
+def warn(path: Path, line: int, msg: str) -> None:
+    warnings.append((path, line, msg))
+
+
+def rel(path: Path) -> str:
+    return str(path.relative_to(ROOT))
+
+
+def parse_frontmatter(text: str):
+    """Return (frontmatter dict or None, error message or None)."""
+    m = re.match(r"\A---\r?\n(.*?)\r?\n---\r?\n", text, re.S)
+    if not m:
+        return None, "missing YAML frontmatter (--- ... ---) at the top of the file"
+    block = m.group(1)
+    try:
+        import yaml  # available on GitHub runners; optional locally
+        data = yaml.safe_load(block)
+        if not isinstance(data, dict):
+            return None, "frontmatter is not a YAML mapping"
+        return data, None
+    except ImportError:
+        pass
+    # Minimal fallback: top-level `key: value` and `key: >`/`key: |` folded blocks.
+    data: dict[str, str] = {}
+    key = None
+    folded: list[str] = []
+    for line in block.splitlines():
+        if line[:1] not in (" ", "\t") and ":" in line:
+            if key is not None:
+                data[key] = " ".join(folded).strip()
+            key, _, val = line.partition(":")
+            key = key.strip()
+            val = val.strip()
+            if val in (">", "|", ">-", "|-"):
+                folded = []
+            else:
+                data[key] = val
+                key = None
+        elif key is not None and line.strip():
+            folded.append(line.strip())
+    if key is not None:
+        data[key] = " ".join(folded).strip()
+    return data, None
+
+
+def code_tokens(path: Path):
+    """Yield (line_number, token) for tokens inside inline code spans and fenced blocks."""
+    in_fence = False
+    for lineno, line in enumerate(path.read_text().splitlines(), start=1):
+        stripped = line.strip()
+        if stripped.startswith("```"):
+            in_fence = not in_fence
+            continue
+        chunks = [line] if in_fence else re.findall(r"`([^`]+)`", line)
+        for chunk in chunks:
+            for tok in TOKEN_SPLIT.split(chunk):
+                tok = tok.rstrip(".,:;")  # rstrip only: `.claude/...` keeps its leading dot
+                if tok:
+                    yield lineno, tok
+
+
+def is_placeholder(tok: str) -> bool:
+    low = tok.lower()
+    return any(marker in low for marker in PLACEHOLDER_MARKERS)
+
+
+def path_exists(tok: str, doc: Path) -> bool:
+    tok = tok.rstrip("/")
+    bases = [ROOT]
+    if doc.parent != ROOT:
+        bases.insert(0, doc.parent)  # skill docs may use paths relative to the skill dir
+    for base in bases:
+        if "*" in tok:
+            try:
+                if any(base.glob(tok)):
+                    return True
+            except (ValueError, NotImplementedError):
+                pass
+        elif (base / tok).exists():
+            return True
+    return False
+
+
+def check_paths(doc: Path) -> None:
+    seen: set[str] = set()
+    for lineno, tok in code_tokens(doc):
+        if tok in seen or is_placeholder(tok):
+            continue
+        if not (tok in TOP_LEVEL_FILES or tok.startswith(DIR_PREFIXES)):
+            continue
+        seen.add(tok)
+        if not path_exists(tok, doc):
+            error(doc, lineno, f"references `{tok}`, which does not exist in the repository")
+
+
+def check_project_flags(doc: Path) -> None:
+    text = doc.read_text()
+    for lineno, line in enumerate(text.splitlines(), start=1):
+        for proj in PROJECT_FLAG_RE.findall(line):
+            if proj in (".", "@.") or is_placeholder(proj):
+                continue
+            if not (ROOT / proj / "Project.toml").is_file():
+                error(doc, lineno, f"`--project={proj}` but {proj}/Project.toml does not exist")
+
+
+def check_size(doc: Path, max_lines: int, max_bytes: int | None = None) -> None:
+    n_lines = len(doc.read_text().splitlines())
+    n_bytes = doc.stat().st_size
+    if n_lines > max_lines:
+        error(doc, 1, f"{n_lines} lines exceeds the {max_lines}-line budget; move detail into a skill reference or delete it")
+    elif n_lines > 0.8 * max_lines:
+        warn(doc, 1, f"{n_lines} lines is close to the {max_lines}-line budget")
+    if max_bytes is not None and n_bytes > max_bytes:
+        error(doc, 1, f"{n_bytes} bytes exceeds the {max_bytes}-byte budget")
+
+
+def check_skill(skill_md: Path, seen_names: dict[str, Path]) -> None:
+    text = skill_md.read_text()
+    fm, err = parse_frontmatter(text)
+    if err:
+        error(skill_md, 1, err)
+        return
+
+    name = fm.get("name")
+    if not isinstance(name, str) or not name:
+        error(skill_md, 1, "frontmatter is missing a `name`")
+    else:
+        if len(name) > LIMITS["skill_name_len"]:
+            error(skill_md, 1, f"skill name is {len(name)} chars (max {LIMITS['skill_name_len']})")
+        if not SKILL_NAME_RE.match(name):
+            error(skill_md, 1, f"skill name `{name}` must be lowercase letters, digits, and hyphens")
+        if name != skill_md.parent.name:
+            error(skill_md, 1, f"skill name `{name}` does not match its directory `{skill_md.parent.name}`")
+        if name in seen_names:
+            error(skill_md, 1, f"skill name `{name}` is also used by {rel(seen_names[name])}")
+        else:
+            seen_names[name] = skill_md
+
+    desc = fm.get("description")
+    if not isinstance(desc, str) or not desc.strip():
+        error(skill_md, 1, "frontmatter is missing a `description`")
+    else:
+        desc = " ".join(desc.split())
+        if len(desc) > LIMITS["skill_desc_len"]:
+            error(skill_md, 1, f"description is {len(desc)} chars (max {LIMITS['skill_desc_len']})")
+        if len(desc) < LIMITS["skill_desc_min"]:
+            error(skill_md, 1, "description is too short to say what the skill does and when to use it")
+
+    for key in fm:
+        if key not in KNOWN_SKILL_KEYS:
+            warn(skill_md, 1, f"unexpected frontmatter key `{key}`")
+
+    check_size(skill_md, LIMITS["skill_lines"])
+
+    for ref in skill_md.parent.rglob("*.md"):
+        if ref != skill_md:
+            check_size(ref, LIMITS["reference_lines"])
+
+
+def check_duplicates(docs: list[Path]) -> None:
+    lines_by_content: dict[str, list[tuple[Path, int]]] = {}
+    for doc in docs:
+        for lineno, line in enumerate(doc.read_text().splitlines(), start=1):
+            norm = " ".join(line.split())
+            if len(norm) < LIMITS["dup_min_len"] or norm.startswith(("#", "```", "-->", "<!--")):
+                continue
+            lines_by_content.setdefault(norm, []).append((doc, lineno))
+    for norm, hits in lines_by_content.items():
+        files = {doc for doc, _ in hits}
+        if len(files) > 1:
+            doc, lineno = hits[0]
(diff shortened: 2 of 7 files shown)


cossio:
@claude review

.github/scripts/lint_agent_docs.py
```diff
@@ -0,0 +1,284 @@
...
+
+def parse_frontmatter(text: str):
+    """Return (frontmatter dict or None, error message or None)."""
+    m = re.match(r"\A---\r?\n(.*?)\r?\n---\r?\n", text, re.S)
+    if not m:
+        return None, "missing YAML frontmatter (--- ... ---) at the top of the file"
+    block = m.group(1)
+    try:
+        import yaml  # available on GitHub runners; optional locally
+        data = yaml.safe_load(block)
```
cossio: **Bug (confirmed):** malformed YAML frontmatter crashes the linter with a raw traceback instead of a clean lint error. `yaml.safe_load` can raise `yaml.YAMLError`, which is not caught (only `ImportError` is). Reproduced with:

```yaml
---
name: [unclosed
description: "..."
---
```

CI still fails (non-zero exit), but with an unhelpful traceback and no `::error` annotation pointing at the offending file. Suggest:

```python
    try:
        import yaml  # available on GitHub runners; optional locally
    except ImportError:
        yaml = None
    if yaml is not None:
        try:
            data = yaml.safe_load(block)
        except yaml.YAMLError as e:
            return None, f"frontmatter is not valid YAML: {e}"
        if not isinstance(data, dict):
            return None, "frontmatter is not a YAML mapping"
        return data, None
```

---

.github/scripts/lint_agent_docs.py
```diff
@@ -0,0 +1,284 @@
...
+    in_fence = False
+    for lineno, line in enumerate(path.read_text().splitlines(), start=1):
+        stripped = line.strip()
+        if stripped.startswith("```"):
+            in_fence = not in_fence
+            continue
+        chunks = [line] if in_fence else re.findall(r"`([^`]+)`", line)
+        for chunk in chunks:
+            for tok in TOKEN_SPLIT.split(chunk):
+                tok = tok.rstrip(".,:;")  # rstrip only: `.claude/...` keeps its leading dot
```
cossio, line 134: **False positive (confirmed):** `file.jl:42`-style line references get flagged as nonexistent paths. `rstrip(".,:;")` strips a trailing colon but not a `:<digits>` suffix, so a doc that writes `` `src/rbm.jl:42` `` produces `ERROR: references src/rbm.jl:42, which does not exist`. No current doc does this, but it's a natural way to cite code and would break CI the first time someone uses it. Suggest stripping the suffix before the existence check:

```python
tok = re.sub(r":\d+(-\d+)?$", "", tok)
```

---

.github/workflows/agent-docs.yml
```diff
@@ -0,0 +1,114 @@
...
+          else
+            echo "changed=false" >> "$GITHUB_OUTPUT"
+            echo "No agent docs changed; skipping Claude review."
+          fi
+      - name: Note when the API key is unavailable
+        if: steps.changed.outputs.changed == 'true' && env.HAS_KEY != 'true'
+        run: echo "::notice::ANTHROPIC_API_KEY is not available (fork PR or missing secret); skipping Claude review."
+      - name: Review agent docs with Claude
+        if: steps.changed.outputs.changed == 'true' && env.HAS_KEY == 'true'
+        uses: anthropics/claude-code-action@v1
```
cossio: **Heads-up: this step has never executed.** On this PR's own CI run the job logs show `HAS_KEY: false` — the `ANTHROPIC_API_KEY` secret is not configured in this repo, so the Claude review skipped (the graceful-skip path works, which is nice, but it means the entire semantic layer is unproven). Two things before relying on it:

1. Add the `ANTHROPIC_API_KEY` repo secret and push a doc-touching commit to see one real run.
2. When it does run, verify the action's permission requirements: `claude-code-action@v1` typically wants `id-token: write` (and/or a `github_token`) in addition to `contents: read`. With only `contents: read` at the workflow level, the first real run may fail at action setup. Untested paths in CI have a way of failing exactly when you first need them.

Also consider adding `timeout-minutes:` to this job — a stuck 40-turn agentic run otherwise bills runner time up to the 6-hour default.

---

.github/workflows/agent-docs.yml
```diff
@@ -0,0 +1,114 @@
...
+    env:
+      HAS_KEY: ${{ secrets.ANTHROPIC_API_KEY != '' }}
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0
+        with:
+          fetch-depth: 0
+      - name: Check whether agent docs changed
+        id: changed
+        run: |
+          if git diff --name-only "origin/${{ github.base_ref }}...HEAD" \
```
cossio: **Hardening (minor):** interpolating `${{ github.base_ref }}` directly into a `run:` script is the template-injection pattern GitHub's security docs warn about. Risk is low here (the base ref must be an existing branch in this repo), but the default environment variable is already available and avoids the pattern entirely:

```bash
if git diff --name-only "origin/${GITHUB_BASE_REF}...HEAD" \
```

Separately: the changed-files filter doesn't include `.github/workflows/agent-docs.yml` or `.github/scripts/lint_agent_docs.py` themselves, so PRs that only change the review prompt or linter won't get the Claude review exercised. That may be intentional (cost), just flagging it.

---

.github/workflows/agent-docs.yml
```diff
@@ -0,0 +1,114 @@
...
+      - name: Enforce review verdict
+        if: steps.changed.outputs.changed == 'true' && env.HAS_KEY == 'true'
+        run: |
+          if [ ! -f agent-docs-review.md ]; then
+            echo "::warning::Claude did not produce agent-docs-review.md; treating the review as inconclusive."
+            exit 0
+          fi
+          cat agent-docs-review.md >> "$GITHUB_STEP_SUMMARY"
+          cat agent-docs-review.md
+          if grep -q '^VERDICT: FAIL' agent-docs-review.md; then
```
cossio: **Edge case (minor):** `grep -q '^VERDICT: FAIL'` matches *any* line starting with `VERDICT: FAIL`, not just the final line the prompt asks for. If the review body ever quotes or explains the verdict format at the start of a line, a PASS review gets treated as FAIL. `tail -n 1 agent-docs-review.md | grep -qx 'VERDICT: FAIL'` would match the contract in the prompt exactly (maybe with a fallback warning when the last line is neither PASS nor FAIL).

---

.agents/skills/restricted-boltzmann-machines-jl/references/architecture.md
```diff
@@ -21,9 +21,8 @@ Read this file when you need the package map or repo-specific edit guidance.
 - `src/rbm.jl` and `src/rbms/`
   Core RBM types plus convenience constructors and variants.
 - `src/train/`
-  Initialization, PCD, gradients, and minibatch-related training helpers.
-  Note that `src/train/minibatches.jl` exists but is not currently included by
-  the root module.
+  Initialization, PCD, gradients, and minibatch iteration helpers
```
cossio: **Nit:** the fix is accurate (verified `src/train/` now contains `infinite_minibatches.jl` and `minibatches.jl` is gone), but the enumeration omits UCD — `src/train/ucd.jl` exists and is included by the root module (`src/RestrictedBoltzmannMachines.jl:47`). Since this sentence lists what lives in `src/train/`, consider "Initialization, PCD, UCD, gradients, and minibatch iteration helpers".

---

cossio (commented):
Reviewed the linter, the workflow, and the doc fixes, exercising the code rather than just reading it. Overall this is solid: the design (deterministic layer + semantic layer, graceful skip on missing key) is sound, and the doc corrections are accurate.

**What I verified locally on the PR branch:**
- `lint_agent_docs.py` runs clean on the repo as-is (5 docs checked, 0 warnings, exit 0).
- Stale-path detection works: a fabricated `.claude/CLAUDE.md` reference, a deleted `src/train/minibatches.jl` reference, and a bogus `--project=nonexistent` flag are all caught with correct `::error` annotations. Notably, the pre-fix `AGENTS.md` reference to `.claude/CLAUDE.md` would have failed this very linter, so the doc fixes in this PR are load-bearing.
- The no-PyYAML fallback frontmatter parser correctly parses the real SKILL.md (name + folded description).
- The doc claims check out: `src/train/` contains `infinite_minibatches.jl` (not `minibatches.jl`), and `CLAUDE.md` exists at the root while `.claude/CLAUDE.md` does not.

**Findings (inline comments have details):**
1. Confirmed bug: malformed YAML frontmatter crashes the linter with a traceback (`yaml.YAMLError` uncaught) instead of a clean lint error.
2. Confirmed false positive: `src/rbm.jl:42`-style file:line citations are flagged as nonexistent paths.
3. **The Claude review layer has never run**: this PR's own CI shows `HAS_KEY: false` — the `ANTHROPIC_API_KEY` secret isn't configured, so the `claude-code-action` invocation (inputs, permissions, verdict flow) is untested. Worth adding the secret and validating one real run; `@v1` typically also needs `id-token: write`.
4. Minor: `${{ github.base_ref }}` interpolated into a `run:` script (use `$GITHUB_BASE_REF`), verdict grep matches any line rather than the final line, no `timeout-minutes` on the review job, and the architecture.md blurb omits `src/train/ucd.jl`.

None of these block merging the lint layer, which works today. Item 3 is the one to close out before trusting the semantic layer.

---

cossio:
Addressed all review feedback in 32037b0:

**Linter**
- Malformed frontmatter now produces a clean `::error` annotation (`yaml.YAMLError` caught) instead of a traceback — verified against the `name: [unclosed` repro.
- `src/rbm.jl:42` / `:10-20` style citations are stripped to the path before the existence check, so they no longer false-positive (a dead `src/gone.jl:7` is still caught).
- Added `from __future__ import annotations` for Python < 3.10 runners.

**Workflow**
- Pinned `anthropics/claude-code-action` to commit `f1bd27c` (v1.0.173); dependabot's github-actions config will keep it updated.
- Replaced `${{ github.base_ref }}` interpolation with `$GITHUB_BASE_REF`, diffing against `origin/$GITHUB_BASE_REF` (guaranteed present with `fetch-depth: 0`).
- Review job now has `id-token: write` + `pull-requests: read` and `timeout-minutes: 30`.
- Verdict is enforced on the final line only, with an inconclusive-warning fallback when the last line is neither PASS nor FAIL.
- The changed-files filter now includes the workflow and linter themselves, so prompt/linter changes exercise the review.

**Docs** — `architecture.md` now lists UCD in the `src/train/` enumeration.

**Remaining, needs repo admin:** the semantic layer stays unexercised until the `ANTHROPIC_API_KEY` secret is added (Settings → Secrets and variables → Actions). Once it's set, this PR's own docs changes will trigger a real run on the next push, which is the validation run the review asked for.

---

cossio (commented):
Re-reviewed 32037b0, verifying each fix locally rather than by inspection:

- **YAML crash fixed:** malformed frontmatter now produces a single clean `::error file=...` annotation (message flattened to one line, so the GitHub annotation isn't truncated) instead of a traceback.
- **file:line false positive fixed:** `src/rbm.jl:42` and range citations like `:10-20` pass, while a stale path *with* a line suffix (`src/train/gone.jl:5`) is still correctly reported — the suffix is stripped before the existence check, not used to skip it.
- Repo still lints clean (5 docs, 0 warnings), and the 08:14 CI run is green for both agent-docs jobs, which also proves the pinned `claude-code-action` SHA resolves at job setup.
- Workflow hardening all landed: SHA pin, `$GITHUB_BASE_REF`, `id-token: write` + `pull-requests: read`, `timeout-minutes: 30`, final-line verdict parsing with an inconclusive-warning fallback, and the changed-files filter now covers the workflow/linter themselves.

The one open item is not fixable from a commit: the `ANTHROPIC_API_KEY` repo secret is still unset (`HAS_KEY: false` on the 08:14 run), so the semantic review layer remains unexercised. I'm leaving that single thread unresolved as the reminder; everything else is resolved.

---

cossio (commented):
Reviewed 5e527b4. Dropping the never-exercisable Claude-review job rather than shipping a permanently-skipped code path is the right call, and the removal is clean — verified locally at this commit: the linter still passes (5 docs, 0 warnings), the workflow YAML parses, and no dangling references to the removed job remain (the linter docstring was updated consistently). I've resolved the last open thread, which this commit moots.

CI on this head: lint, Julia 1.10, CodeQL, and both codecov checks are green; Julia-latest and the docs build are still running. The codecov project dip (100% → 99.96%, one line) is unrelated to this PR — it touches no Julia source — and sits well within the 1% threshold.

One follow-up: the PR description still advertises the two-layer system (`claude-review` job, ANTHROPIC_API_KEY skip behavior, semantic-review implementation notes). Worth trimming it to describe the lint-only design before merge, since the description is what future archaeology will read.

---

.github/workflows/claude-pr-review.yml
```diff
@@ -0,0 +1,83 @@
...
+          fetch-depth: 0
+      - name: Review pull request with Claude
+        if: env.HAS_KEY == 'true'
+        uses: anthropics/claude-code-action@f1bd27ca5b54584506e40e17884d90bdaaa1a9b3 # v1.0.173
+        with:
+          claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
+          claude_args: |
+            --model claude-sonnet-5
+            --max-turns 60
+            --allowedTools "Read,Grep,Glob,LS,Bash(gh pr diff:*),Bash(gh pr view:*),Bash(gh api graphql:*),Bash(git log:*),Bash(git diff:*),Bash(git show:*),mcp__github_inline_comment__create_inline_comment"
```
cossio: **Note (accepted-risk callout):** `Bash(gh api graphql:*)` is broader than the thread-reconciliation it's there for — it permits any GraphQL query or mutation the job's app token can perform (with `pull-requests: write`, that includes editing/dismissing reviews, closing PRs, etc.). It's bounded by the job's permission set, so this is acceptable, but it's the widest grant in the list and worth being aware of. If you ever want to narrow it, `resolveReviewThread` calls could go through a dedicated step instead of the agent's Bash.

---

.github/workflows/claude.yml
```diff
@@ -0,0 +1,53 @@
...
+    types: [created]
+  pull_request_review_comment:
+    types: [created]
+  pull_request_review:
+    types: [submitted]
+  issues:
+    types: [opened]
+permissions:
+  contents: read
+jobs:
```
cossio, line 26: **Minor:** no `concurrency` group here, so several near-simultaneous `@claude` mentions (e.g. replies across multiple review threads) run as parallel 45-minute jobs, possibly pushing to the same branch concurrently. A group keyed on the issue/PR number with `cancel-in-progress: false` (queueing, not cancelling — you don't want a half-finished fix commit) would serialize them:

```yaml
concurrency:
  group: claude-${{ github.event.issue.number || github.event.pull_request.number }}
```

---

cossio (commented):
Reviewed 44a6853 and 31b2c9c together, including digging into the CI logs of the re-enabled review job. The headline finding changes how this PR should be judged:

**The Claude review jobs cannot be exercised from this PR — by design of the action itself.** The `CLAUDE_CODE_OAUTH_TOKEN` secret is now configured and working (job logs show `HAS_KEY: true` and a successful OIDC→app-token exchange), but `claude-code-action` then self-skips with:

> *"Workflow validation failed. The workflow file must exist and have identical content to the version on the repository's default branch. … your workflow will begin working once you merge your PR."*

This is the action's anti-tamper protection (it stops a PR from editing the prompt/allowedTools and running the modified version with repo credentials), and it means:

1. **Nothing on this PR can prove the reviews work end-to-end.** Both the agent-docs review and the new `Claude review` job took this skip path on the latest heads (~30s "success" runs). The real test happens on the first doc- or code-touching PR *after* merge — worth watching that one.
2. The `agent-docs.yml` changed-files filter including the workflow file itself (my earlier suggestion) delivers less than intended: any PR editing the workflow triggers the review job, but the action then always self-skips, yielding an "inconclusive" warning run. Harmless, just no longer load-bearing.
3. The verdict fallback (warn-don't-fail when `agent-docs-review.md` is missing) earned its keep here — without it, this PR's own CI would be red for an unfixable reason.

On the two new workflows themselves: `claude-pr-review.yml` is well-constructed — advisory-only, thread reconciliation before posting to avoid duplicates, concurrency-cancel on supersede, draft skip, SHA-pinned action, bounded turns/timeout, fork-safe skip. `claude.yml`'s trigger gating on OWNER/MEMBER/COLLABORATOR association is the right guard for a public repo given its `contents: write`. Both YAML-parse clean and reuse the established skip pattern correctly. Two minor inline notes.

One scope observation: this PR now ships three Claude workflows plus the linter — well beyond the "agent docs CI" in the title/description. Fine to merge as-is, but the description rewrite flagged earlier is now even more warranted, and a split (agent-docs CI vs. general Claude automation) would have made the history cleaner.

---

cossio (commented):
Reviewed 424d40f. Good consolidation — one agentic run per PR instead of two, and the removal is complete: no dangling references to the verdict machinery or the removed job anywhere under `.github/` (checked), all three workflows YAML-parse, the linter still passes, and `agent-docs.yml` is back to exactly the lint-only version that already ran green (byte-identical to its 5e527b4 state apart from the three-line pointer comment).

One trade-off to make consciously, since the commit message doesn't dwell on it: the agent-docs semantic review went from **blocking** (the `VERDICT: FAIL` gate) to **advisory** (inline comments that can't fail the job). I think that's the right call — the deterministic linter remains the enforcement layer, the blocking variant could never be exercised from a PR that touches its own workflow anyway, and inline comments are more actionable than a step-summary file — but it does mean a future PR that introduces, say, a contradiction between CLAUDE.md and the code will merge green unless someone reads the advisory comments.

Standing caveat from the previous review still applies: none of this can execute on this PR (the action's workflow-validation skip), so the first doc- or code-touching PR after merge is the real end-to-end test of the unified review.

---

cossio:
The refreshed description is much closer, but it now *under*-describes the PR: it stops at the lint-only state (5e527b4) and says the semantic review "was removed" and "remains a manual/agent-assisted task." As of 424d40f that's no longer accurate — the diff also ships **`.github/workflows/claude-pr-review.yml`** (advisory Claude review of every PR, which now carries the agent-docs semantic dimension, authenticated with `CLAUDE_CODE_OAUTH_TOKEN`) and **`.github/workflows/claude.yml`** (@claude tag mode, gated to owner/members/collaborators). Suggest a "Changes" bullet for each and dropping the "removed since it requires an ANTHROPIC_API_KEY secret" note in favor of the actual outcome: the semantic review lives in the unified PR review, and only starts executing after this PR merges (the action self-skips on PRs whose workflow files differ from the default branch).

Meanwhile CI on 424d40f is green across the board (both Julia versions, lint, CodeQL, codecov); only the docs build is still running.

---

cossio (commented):
Reviewed 5c271c2. The verdict mechanism is well-specified — the three-way rubric puts the bar for `--request-changes` in the right place ("confirmed, consequential — never for uncertainty or taste"), the fresh-verdict-supersedes rule handles re-runs correctly (GitHub counts each reviewer's latest review, so a post-fix approval clears the bot's earlier request-changes), and the permission-rejection fallback to a comment is a sensible safety net. YAML parses; the header comment accurately reflects the new behavior.

Three observations, none blocking:

1. **`Bash(gh pr review:*)` adds no new capability in practice** — the pre-existing `Bash(gh api graphql:*)` grant could already submit reviews via the `addPullRequestReview` mutation. So this doesn't widen the security surface; it just makes the intended path explicit.
2. **The approval-permission note may be moot in a good way:** the "Allow GitHub Actions to create and approve pull requests" setting governs the workflow's own `GITHUB_TOKEN`, but `claude-code-action` submits reviews with the exchanged Claude GitHub App token, which isn't gated by that setting. Either way the fallback covers rejection, so nothing to change — just don't be surprised if approval works even where the setting is off.
3. **Forward-looking:** if this repo ever adds branch protection requiring an approving review, the bot's approval will count toward it — i.e., an agent approval can satisfy the human-review requirement. Intended for a solo-maintainer repo, but worth remembering when the rule is configured.

Standing caveat unchanged: none of this executes on this PR (workflow-validation self-skip); the first PR after merge is the live test of the whole pipeline, now including the verdict step.

---

cossio (commented):
Reviewed 895f8e8. Good closing of the loop — the merge boundary ("never merge or enable auto-merge; merging is reserved for the repo owner") is now stated in both places that matter: CLAUDE.md for PR-authoring agents and the tag-mode system prompt for the workflow that actually holds `contents: write`. The new CLAUDE.md section passes the agent-docs linter (its `.github/workflows/claude-pr-review.yml` reference is path-checked and exists), the claims in it match the actual workflow behavior (re-runs on every push, fresh verdict supersedes), and `claude.yml` still YAML-parses with the appended prompt intact as one quoted argument.

One thing to eyeball on the first live tag-mode run: `--append-system-prompt "…"` sits inside the `claude_args` block scalar, so the quoted multi-word string must survive the action's argument splitting. It should (the action does shell-style parsing of `claude_args`), but since nothing on this PR can execute pre-merge, that's one more item for the first-real-run checklist rather than something verifiable here.

---

.github/workflows/claude-pr-review.yml
cossio: Keeping the broad `Bash(gh api graphql:*)` grant, with reasoning: as of 51aa04e the reviewer's reconcile step is load-bearing on GraphQL for three operations — listing review threads with replies, `resolveReviewThread`, and `addPullRequestReviewThreadReply` (the new reply-or-resolve protocol). A dedicated non-agent step can't do this work, since deciding *which* threads to resolve or answer requires the agent's judgment mid-run. The grant stays bounded by the job's permission set (`contents: read`, `pull-requests: write`), which caps the blast radius at review/PR-metadata operations. Accepting the risk as you suggested; if it ever needs narrowing, a proxy script allowlisting the three mutations would be the route.

---

.github/workflows/claude.yml
cossio: Fixed in 51aa04e — added the concurrency group as suggested, keyed `claude-${{ github.event.issue.number || github.event.pull_request.number }}` with an explicit `cancel-in-progress: false` so concurrent mentions queue rather than kill a half-finished fix commit. One known GitHub limitation worth noting: a concurrency queue holds only one pending run, so if three mentions land near-simultaneously the middle one gets superseded — acceptable here since the surviving run sees the latest state.

---

cossio (commented):
Reviewed 51aa04e. The reply-or-resolve protocol is a good division of labor — the author replies in every thread but never resolves; the reviewer resolves once convinced (by fix or rebuttal) and replies back when not — and it's now encoded consistently in all three places (CLAUDE.md, the reconcile step with the `addPullRequestReviewThreadReply` mutation, and the tag-mode system prompt). The `claude.yml` concurrency group matches what was suggested, with the right `cancel-in-progress: false` choice and a group key that resolves for all four trigger event types (`issue.number` covers `issue_comment` and `issues`; `pull_request.number` covers the two review events). Lint and YAML checks pass at this head.

I've also practiced the protocol on this PR: both replies in my open threads were convincing (the concurrency fix is verified in the diff; the `gh api graphql` breadth is now load-bearing for the reconcile step and stays bounded by job permissions), so I resolved both. **No open threads remain** — from my side this PR is again merge-ready.

One caveat inherited from your own reply, worth restating as the final state: GitHub's concurrency queue holds a single pending run, so of three near-simultaneous `@claude` mentions the middle one is superseded silently — acceptable since the surviving run sees the latest state, but a dropped mention will look like Claude ignoring someone.

---

PR #147, opened 2026-07-17
Return +Inf from Gaussian-Gaussian log_partition for non-normalizable RBMs

### LINKED ISSUE #142: Gaussian-Gaussian log_partition returns a finite value for non-normalizable RBMs
## Summary

`log_partition(::RBM{<:Gaussian,<:Gaussian})` applies the Gaussian integral formula without checking that the joint precision matrix is positive definite. An indefinite matrix can have a positive determinant when it has an even number of negative eigenvalues, so the current block log-determinant is finite even though the energy is unbounded below and the partition function diverges.

Relevant implementation:

- [`log_partition` for Gaussian–Gaussian RBMs](https://github.com/cossio/RestrictedBoltzmannMachines.jl/blob/b1a7c57d/src/rbms/gaussian.jl#L22-L39)

## Reproduction

```julia
import RestrictedBoltzmannMachines as RBMs
using LinearAlgebra: I, eigvals, isposdef

A = Matrix{Float64}(I, 2, 2)
rbm = RBMs.GaussianRBM(
    zeros(2),
    ones(2),
    zeros(2),
    ones(2),
    2A,
)

precision = [A -2A; -2A A]

@show eigvals(precision)
@show isposdef(precision)
@show RBMs.log_partition(rbm)

for t in (1.0, 10.0, 100.0)
    v = t .* [1.0, 0.0]
    h = t .* [1.0, 0.0]
    @show t RBMs.energy(rbm, v, h)
end
```

Observed at `b1a7c57d`:

```text
eigvals(precision) = [-1.0, -1.0, 3.0, 3.0]
isposdef(precision) = false
log_partition(rbm) = 2.5771418441505807
energy at t = 1, 10, 100: -1, -100, -10000
```

Along `v = h = t * e₁`, the energy is `-t²`, so the integral diverges and `Z = ∞`. Returning a finite value can silently corrupt exact likelihood calculations.

## Expected behavior

Validate that the symmetric joint precision matrix is positive definite before applying the finite Gaussian integral formula. For a singular or indefinite matrix, return `+Inf`, matching the extended-real value of the divergent log partition function, rather than a plausible finite number.

A checked Cholesky factorization of the symmetric precision matrix could provide both the validation and a numerically robust determinant/inverse path.

## Acceptance criteria

- The indefinite example above returns `+Inf` instead of a finite value.
- Singular joint precision matrices likewise return `+Inf`.
- Existing results for positive-definite models are preserved.
- Tests include a matrix with an even number of negative eigenvalues, since its determinant alone does not expose the problem.

cossio:
Fixes #142.

## Summary

- validate the symmetric Gaussian-Gaussian joint precision using the model's `abs(γ)` semantics
- return `+Inf` when the precision is singular or indefinite
- reuse a checked Cholesky factor for the positive-definite solve and log-determinant
- document the behavior and add Float32/Float64, boundary, and differentiation regressions

## Root cause

The previous implementation applied a block log-determinant and inverse without checking positive definiteness. An indefinite precision with an even number of negative eigenvalues can have a positive determinant, so `log_partition` could return a plausible finite value even though the energy is unbounded below and the partition function diverges.

## Validation

- `julia --project=test test/rbm.jl`
- `julia --project=. -e 'using Pkg; Pkg.test()'`

All listed checks passed locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file. The format 
 
 ## Unreleased
 
+- Fixed `log_partition` for Gaussian-Gaussian RBMs, which could return a plausible finite value for a non-normalizable model when the indefinite joint precision had a positive determinant. It now validates the `abs(γ)` joint precision with a checked Cholesky factorization and returns `+Inf` for singular or indefinite models ([#142](https://github.com/cossio/RestrictedBoltzmannMachines.jl/issues/142)).
+
 ## 5.8.0
 
 - Added support for weighted data (`wts` keyword) in `pcd!` for `StandardizedRBM`, matching the plain-`RBM` trainer: weighted visible moments, weighted minibatch gradients with the weighted-minibatch bias correction, and weighted updates of the visible/hidden standardization statistics. The `callback` now also receives the minibatch weights `wd` (equal to `nothing` when `wts` is not given), consistent with the plain-`RBM` trainer; callbacks that spell out the full keyword list must accept the new keyword (or, more robustly, take a trailing `_...` slurp).
--- a/src/partition.jl
+++ b/src/partition.jl
@@ -5,6 +5,11 @@ Log-partition of `rbm`, computed by extensive enumeration of visible states
 (except for particular cases such as Gaussian-Gaussian RBM).
 This is exponentially slow for large machines.
 
+For Gaussian-Gaussian RBMs, the exact Gaussian integral is used when the joint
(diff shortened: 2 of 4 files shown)


cossio:
Integrated the latest `master` (including #144) with a normal merge. The only textual conflict was in `CHANGELOG.md`; I preserved both the #142 Gaussian log-partition entry and #144 zero-norm rescaling entry. Validation after reconciliation: `julia --project=test test/rbm.jl` and the full `julia --project=. -e 'using Pkg; Pkg.test()'` suite both pass.

PR #159, opened 2026-07-18
Add downgrade compatibility testing to CI

cossio:
## Summary

- add a dedicated `julia-downgrade-compat` job to the main CI workflow
- resolve the root workspace with minimum direct dependency versions on current Julia, whose Pkg workspace support incorporates the test project's compatibility bounds
- build and run `Pkg.test` from the generated lockfile with re-resolution disabled
- raise the stale lower bounds found by that run: ChainRulesCore to 1.25.2 and test-only Aqua to 0.8.14
- remove the deactivated archived downgrade workflow

## Why

Normal CI exercises current compatible dependency versions, so stale lower compat bounds can go unnoticed. The existing Julia 1.10 job continues to cover the minimum supported Julia runtime; this separate job checks dependency floors across the workspace.

## Validation

- reproduced the pinned downgrade action locally on Julia 1.12.6
- confirmed `Pkg.instantiate()` leaves the generated manifest byte-for-byte unchanged
- ran the full package suite with `allow_reresolve=false` and `force_latest_compatible_version=false`; all tests passed, including HDF5, JLArrays, training, and Aqua
- parsed `.github/workflows/ci.yml` successfully
- `git diff --check`
- checked the branch against the latest `master`; no merge conflict


.github/workflows/ci.yml
```diff
@@ -37,3 +37,27 @@ jobs:
...
+          projects: '.,test'
+          mode: deps
+          julia_version: '1.10'
+      - uses: julia-actions/cache@a45e8fa8be21c18a06b7177052533149e61e9b38
+      - name: Run tests
+        run: |
+          julia --project=test --color=yes -e '
+            import Pkg
+            Pkg.develop(Pkg.PackageSpec(path=pwd()))
+            Pkg.instantiate()
```
cossio: I pulled the actual job log for this run (`Downgrade compat - Julia 1.10`, job 88109760832) to check whether the downgraded floor survives into the test run, since @chatgpt-codex-connector's comment above flags exactly this risk.

It doesn't. The `julia-downgrade-compat` step resolves the merged project with `--min=@deps` and installs `HDF5 v0.16.10` (consistent with root's compat, since the log also shows `HDF5 has compat in both projects, using main project's compat`, i.e. Codex's second finding at line 51 — root's `"0.16, 0.17"` wins over test's `"0.17.2"`). But then this "Run tests" step's `Pkg.instantiate()` (triggered by `Pkg.develop(...)` changing the dependency graph) re-resolves the workspace and immediately upgrades again:

```
HDF5 v0.16.10 ⇒ v0.17.3
HDF5_jll v1.10.5+7 ⇒ v2.1.2+0
```

So the test suite in this job actually ran against HDF5 v0.17.3, not the downgraded floor the action computed. This isn't a hypothetical edge case — it's what happened in this PR's own CI run. The new job currently can't fail on a stale lower bound for any package that's re-resolved away during `instantiate()`, which defeats the stated purpose of the PR ("Normal CI exercises current compatible dependency versions... This adds a separate minimum-compatibility check").

Suggest generating/checking in a downgraded `Manifest.toml` (or workspace manifest) from the `julia-downgrade-compat` step and instantiating with `--offline` (or otherwise preventing `Pkg.instantiate()` from re-resolving) so the test step actually runs against the locked minimum versions.

---

cossio (commented):
Reviewed the diff (new `downgrade` job in `.github/workflows/ci.yml`, removal of the deactivated archived `downgrade.yml` workflow) and the actual CI run for this PR's head commit.

The archival cleanup and job wiring (Julia 1.10, `julia-downgrade-compat` with `mode: deps`, cache, checkout) are reasonable and match the stated intent. However, I verified empirically (via the job logs for this PR's own "Downgrade compat" run) that the job does not currently test what it claims to: the later `Pkg.instantiate()` in the "Run tests" step re-resolves the workspace and silently upgrades packages away from the downgraded floor computed by `julia-downgrade-compat` (concretely, `HDF5 v0.16.10 ⇒ v0.17.3` in this run's own log). This confirms @chatgpt-codex-connector's two inline findings above — left a comment with the log evidence and a suggested fix (lock/reuse the downgraded manifest instead of letting `instantiate()` re-resolve).

As-is, this job will pass regardless of stale lower compat bounds, which is the exact problem the PR is meant to catch. Recommend fixing this (or reconsidering the approach) before merging — submitting as COMMENT since I can't request changes on my own PR.

---
_Generated by Claude Code_

---

.github/workflows/ci.yml
cossio: I can't find the changes described here. `codex/add-downgrade-compat-ci` is still at `1c1803c` (no `cce1d94` commit), there's no new branch or PR titled "Fix downgrade compatibility CI locking" in the repo, and `.github/workflows/ci.yml`/`Project.toml`/`CHANGELOG.md` on this branch are unchanged from before this comment. So as far as I can tell nothing was actually applied — this PR still has the re-resolution bug described above unresolved. Flagging in case the task silently failed to push; happy to re-review once an actual commit/PR shows up.

---

cossio:
@chatgpt-codex-connector this is the second summary in a row (previously `cce1d94`, now `769891b`) claiming a fix was committed and a PR opened, but neither commit exists on `codex/add-downgrade-compat-ci` (still at `1c1803c`) or anywhere else in the repo, and no new PR was created. Please check whether the underlying task is actually failing to push/open a PR — as far as this repo is concerned, no fix has landed and the re-resolution issue on this PR is still open.

---

.github/workflows/ci.yml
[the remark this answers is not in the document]
cossio: Fixed in cc1cd4d4. The job now runs on current Julia and resolves `projects: '.'`, so Pkg's workspace includes the test project. The manual `Pkg.develop`/`Pkg.instantiate` sequence is gone; the job uses `julia-buildpkg` followed by `julia-runtest` with both `allow_reresolve: false` and `force_latest_compatible_version: false`. Locally, `Pkg.instantiate()` left the action-generated manifest checksum unchanged, and the full no-re-resolve `Pkg.test` run passed.

.github/workflows/ci.yml
```diff
@@ -37,3 +37,27 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0
+      - uses: julia-actions/setup-julia@fa02766e078afaaf09b14210362cee14137e6a32
+        with:
+          version: '1.10'
+          arch: x64
+      - uses: julia-actions/julia-downgrade-compat@fab1defb76df9fd672f63c94df73ce131d32e134
+        with:
+          projects: '.,test'
```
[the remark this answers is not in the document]
cossio: Fixed in cc1cd4d4 by resolving the root workspace on current Julia instead of using the action's separate pre-1.12 root/test merge path. Pkg's workspace resolution now incorporates the test member's stricter bounds: the generated lock selected HDF5 0.17.2 (not the root-only 0.16 floor) and Aqua 0.8.14. The same run exposed a genuine root floor issue in ChainRulesCore, so its lower bound is now 1.25.2. The resulting locked test suite passes with re-resolution disabled.

.github/workflows/ci.yml
cossio: You're right—the posted CI log showed that the manual `Pkg.develop`/`Pkg.instantiate` step upgraded HDF5, so that version did not actually test the downgraded lock. The real fix is now pushed as cc1cd4d4: those commands are removed, workspace resolution produces HDF5 0.17.2 directly, and `julia-runtest` runs with re-resolution disabled. Locally the manifest checksum was unchanged across `Pkg.instantiate()`, and the complete locked suite passed.

cossio:
Correction to the earlier automated summaries: `cce1d94`, `769891b`, and `e73060c` were never pushed and should be disregarded. The actual review fix is now present on this PR's branch as [`cc1cd4d4`](https://github.com/cossio/RestrictedBoltzmannMachines.jl/commit/cc1cd4d4b5c50d7673162459e285798b2121e343). I reproduced the pinned downgrade action, verified that instantiation does not change its manifest, and ran the complete suite with re-resolution disabled. The PR is currently mergeable against the latest `master` with no conflicts.

@codex review

cossio (commented):
Reviewed the diff (scheduled PR watch). Overall this is a clean, well-scoped CI change:

- New `downgrade` job in `ci.yml` mirrors the pinned-action style already used by the `test` job (same SHAs for `actions/checkout`, `setup-julia`, `cache`, `julia-buildpkg`, `julia-runtest`), so it's consistent with existing conventions.
- Raising `ChainRulesCore` to `1.25.2` and test-only `Aqua` to `0.8.14` are correctly reflected in `Project.toml`/`test/Project.toml`, and the CHANGELOG entry appropriately covers only the user-facing `ChainRulesCore` bump (not the test-only `Aqua` bump), per this repo's CHANGELOG policy.
- Removing the deactivated `archive (deactivated)/downgrade.yml` workflow in favor of the new in-`ci.yml` job is a reasonable consolidation.

One thing worth double-checking: the old (deactivated) workflow passed `skip: Statistics,Random,LinearAlgebra,Test` to `julia-downgrade-compat`, since those stdlibs carry compat entries (`Statistics = "1.7"`, etc.) tied to the Julia version rather than registry versions. The new job uses `mode: deps` without a `skip` list. If `mode: deps` doesn't already exempt stdlibs from being pinned to their lower compat bound, this could cause spurious resolve failures on some Julia versions even though the code itself is fine. The PR description says this was validated locally on Julia 1.12.6 with all tests passing, and the `Downgrade compat` job is running live in this very PR, so CI should confirm either way — flagging in case that stdlib compat interaction wasn't the exact thing exercised locally.

No functional/source-code correctness issues found; this is infra-only plus two compat-bound bumps.

---

cossio:
Reviewed `cc1cd4d4`. This is a real fix (verified the commit exists and inspected the diff directly, unlike the earlier phantom `cce1d94`/`769891b`/`e73060c` claims from the Codex connector that never actually landed anywhere).

The approach is sound and better than the original: instead of fighting the `julia-downgrade-compat` action's multi-project merge logic (which is what caused the confirmed bug — root's looser `HDF5` compat winning over test's stricter one, then getting silently re-resolved away anyway), it now downgrades only the root project and lets Julia's native `[workspace]` resolution pull in the test member's own compat bounds directly, then locks that down through `julia-runtest`'s `allow_reresolve: false` / `force_latest_compatible_version: false` instead of a manual `Pkg.instantiate()` that could re-resolve. That directly targets the failure mode I found in the CI log (`HDF5 v0.16.10 ⇒ v0.17.3` during the old manual step). The `ChainRulesCore`/`Aqua` floor bumps are a good sign — real stale bounds this run actually caught, which the old (broken) job never could have.

One thing I want to see before calling this settled: the "Downgrade compat - Julia 1" job is still `in_progress` on this commit as I write this. Root's `[compat]` and test's `[compat]` for shared packages (e.g. `Adapt`, `Optimisers`, `SpecialFunctions`) are two independent bounds now being resolved together through workspace resolution rather than the action's merge — if any of those ranges turn out to be disjoint post-downgrade, resolution itself would fail loudly (which is at least safe, unlike the previous silent pass). Will confirm once the run finishes.

---

cossio:
The two live `Downgrade compat - Julia 1` jobs both passed on `cc1cd4d4` (24m48s and 24m59s). The successful log confirms workspace floor resolution and the locked test environment: HDF5 0.17.2, ChainRulesCore 1.25.2, and Aqua 0.8.14. It also answers the stdlib question: Statistics, LinearAlgebra, Random, and Test remained at the runtime-provided stdlib versions, so no `skip` list was needed. The job ended with `RestrictedBoltzmannMachines tests passed`.

PR #165, opened 2026-07-19
Adopt Runic formatting in CI

cossio:
## Summary

- apply Runic 1.7 formatting to all tracked Julia files in an isolated mechanical commit
- add a CI formatting check using the official `fredrikekre/runic-action`, pinned to the v1.4.0 commit SHA
- deliberately pin Runic to the `1.7` minor series so formatting-output changes are explicit upgrades
- record the exact formatting commit in `.git-blame-ignore-revs`

## Scope

The repository-wide formatter migration is isolated in commit `8044b179b339b41ca526ee24520c8bedf4941ece`. CI and blame-ignore metadata were added in the subsequent tooling commit `bff14c968cdfcae63a6a00e4ddcc00a9291cb561`.

Review of `src/util/truncated_normal.jl` led to one deliberate, behavior-preserving hand rewrite after the mechanical migration. `randnt` now uses `local r = a`, keeps the two rejection samplers as `while true` loops that break on acceptance, evaluates the positive sampler only for `a < Inf`, and returns `r` once after the conditional. Thus `Inf` and `NaN` return unchanged without consuming RNG draws. A deterministic test RNG now verifies the exact draw sequences for both rejection branches and the no-draw non-finite cases.

The branch also merges current `master` after PR #164. Its new Gaussian pseudolikelihood implementation and tests are preserved; the affected Julia files were formatted with the pinned Runic version as part of the merge resolution.

No public behavior, API, or dependency changes are intended. No CHANGELOG entry is included for this tooling-only work. No `# runic: off` / `# runic: on` opt-outs were needed, and Literate `#hide` directives remain intact.

## Validation

- `Runic 1.7.0 --check --diff` passed all 86 tracked `.jl` files
- `julia --project=test test/truncnorm.jl` passed 39/39 assertions, including deterministic RNG draw-sequence coverage
- `julia --project=test test/pseudolikelihood.jl` passed 76/76 assertions after integrating current `master`
- `julia --project=test test/jlarrays.jl` passed with one pre-existing broken Potts test and no failures
- `julia --project=. -e 'using Pkg; Pkg.test()'` passed the full suite, including HDF5, JLArrays, and Aqua, with the same one pre-existing broken Potts test and no failures
- `import RestrictedBoltzmannMachines as RBMs` passed on Julia 1.12.6
- `git diff --check` passed
- `.git-blame-ignore-revs` was validated with `git blame --ignore-revs-file`
- local documentation generation successfully converted all Literate sources and executed the MNIST and AIS pages without formatting/directive errors; the full documentation build is covered by CI


.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Apply Runic formatting
+8044b179b339b41ca526ee24520c8bedf4941ece
```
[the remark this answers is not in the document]
cossio: `8044b179b339b41ca526ee24520c8bedf4941ece` is intentionally the isolated formatting commit and is an ancestor of the PR head (`bff14c96`). I fetched GitHub’s current synthetic merge ref (`4fffc9d1`), whose parents are `e77b1b24` and `bff14c96`; `git merge-base --is-ancestor 8044b179 4fffc9d1` exits 0. I also validated the file with `git blame --ignore-revs-file .git-blame-ignore-revs`. The repository uses merge commits, so the formatting commit remains in history and is the correct revision to ignore. The cited `93adcff` is not a repository commit and does not match the review’s recorded head (`bff14c96`), so no change is needed.

cossio (commented):
No blocking issues found.

Spot-checked the isolated formatting commit (`8044b179`) across `src/`, `test/`, `docs/src/literate/`, and `repl/`: all changes reduce to whitespace/style (kwarg `=` spacing, `for x = r` → `for x in r`, operator spacing, multi-line call/matrix reflow) or Runic's documented explicit-return insertion (adds `return` only in front of an already-last expression, so it's a behavior no-op) and numeric literal canonicalization (`1e-6` → `1.0e-6`, value-identical). I specifically checked the matrix-literal reflows in `src/rbms/gaussian.jl` and `test/rbm.jl` for the classic Runic/JuliaFormatter pitfall of touching spacing around unary `-` inside `[...]` literals (which would silently change element grouping) — the spacing there is preserved exactly, so no semantic change.

CI addition (`.github/workflows/format.yml`) looks sound: triggers are `push`/`pull_request` (no `pull_request_target`), `permissions: contents: read` only, and all three actions (`actions/checkout`, `julia-actions/setup-julia`, `fredrikekre/runic-action`) are pinned to full 40-character commit SHAs rather than mutable tags.

On the `.git-blame-ignore-revs` thread: agree with @cossio's resolution of the Codex bot's comment — `8044b179` is the correct isolated formatting commit and is an ancestor of the PR head, so the blame-ignore entry is correct as written; that thread is already resolved and I have nothing to add.

---

src/util/truncated_normal.jl
```diff
@@ -19,7 +19,7 @@ randnt(rng::AbstractRNG, a::BigFloat) = randnt(rng, Float64(a))
 randnt(a::Real) = randnt(default_rng(), a)
 
 function randnt(rng::AbstractRNG, a::Base.IEEEFloat)
-    if a ≤ 0
+    return if a ≤ 0
```
cossio: Why is this `return` needed here?

cossio: It's not functionally needed — before this PR the function ended with a bare `if a ≤ 0 ... else ... end` and returned its value implicitly, since it was already the last expression in the function body. Runic's formatter adds an explicit `return` in front of an already-last `if`/`else` expression as a style convention (make the return value visually explicit), which is a documented no-op transformation — it doesn't change control flow or the returned value, since the `if` was already in tail position. Same behavior either way; happy to revert this one spot if you'd prefer to keep it implicit, but the current form matches how Runic formats every other function-tail `if` in this PR.

---

.git-blame-ignore-revs
cossio, line 1: Explain why this file is needed?

cossio: `.git-blame-ignore-revs` is the standard mechanism `git blame` uses (via `git blame --ignore-revs-file .git-blame-ignore-revs`, or once `git config blame.ignoreRevsFile .git-blame-ignore-revs` is set locally) to skip specific commits when attributing a line's authorship. GitHub's web blame view also auto-detects this exact filename at the repo root and applies it automatically, with no per-user config needed.

It's needed here specifically because this PR's first commit (`8044b179`) reformats all 86 tracked `.jl` files with Runic in one mechanical, repository-wide pass. Without this file, `git blame` would attribute every reformatted line to that commit/date instead of the commit that last changed its actual content — making blame useless for tracking down when a piece of logic actually changed. Listing `8044b179` here tells `git blame` (and GitHub) to look past that commit and keep attributing lines to their real last substantive change.

---

src/util/truncated_normal.jl
cossio: But can we put the `return` statements inside the `if` ? That is, I prefer:

```julia
if cond
  ... 
  return something
else
  ...
  return something
end
```

over

```julia
return if cond
  ... 
  something
else
  ...
  something
end
```

Is my preferred version compatible with Runic?

cossio: Checked upstream: this isn't configurable in Runic 1.7 today. Runic unconditionally hoists a single `return` in front of an already-last `if`/`try` when no branch already has an explicit `return`, rather than duplicating `return` into each branch. Recursing into branches to add per-branch returns instead is an open, unimplemented feature request — [fredrikekre/Runic.jl#52](https://github.com/fredrikekre/Runic.jl/issues/52) — with no target release.

Since this PR adds Runic as a CI-enforced formatter (`format.yml` running `--check`), hand-editing individual spots to your preferred per-branch style would just get flagged as a formatting violation the next time CI (or anyone) runs Runic — it isn't a one-off fix I can make stick.

Given that, the options are:
1. Accept this Runic convention repo-wide (what the PR currently does) and revisit if/when Runic.jl#52 lands.
2. Carve out per-instance exceptions with `# runic: off` / `# runic: on` around such tail-`if`s — the PR description explicitly avoided this to keep the pass fully mechanical, so adopting it would be a deliberate scope change worth calling out in the PR body.
3. Hold off on adopting Runic (or pin an even older version) if this convention is a blocker.

This is a repo-wide style/tooling call rather than something to silently patch in one file, so I'll leave it for you to pick a direction rather than pushing a change.

---

cossio: Correction to my last reply — yes, it is compatible, and I stated the rule backwards. Runic's actual rule is: it hoists a single `return` in front of a tail `if`/`try` **only when no branch already contains an explicit `return`**; if a branch already has one, Runic leaves the block alone. So if you rewrite `randnt` (and similarly elsewhere) to your preferred per-branch-`return` style, `Runic --check` will accept it unchanged — it won't collapse it back into `return if ... end`.

The catch is Runic won't do this rewrite *for* you — the mechanical pass in this PR only added the hoisted form because no branch had a `return` yet. Getting your preferred style means manually adding `return` to every branch of every affected tail-`if`/`try` (there are quite a few across the 86-file diff, since this pattern is common), which is real editing work, not something `--check` will fail you into doing. So the actual options are still yours to weigh: accept the hoisted form as this PR produced it, or do the (larger) manual per-branch-return pass yourself/have it done as a follow-up.

---

cossio (commented):
Reviewed this as a formatting-only PR per REVIEW.md (no comments on style/formatting itself). Focused on the non-mechanical parts and on ruling out any semantic change hiding in the reformatting commit:

- `.github/workflows/format.yml`: third-party actions pinned to full commit SHAs (with version comments), `permissions: contents: read` is minimal. No security issues.
- `.git-blame-ignore-revs`: correctly points at the isolated formatting commit `8044b179b`.
- Spot-checked the mechanical commit for operator-spacing changes that could alter semantics (`.^2` → `.^ 2`, `for a = -10:10` → `for a in -10:10`, `θ^2/abs(2γ)` spacing, etc.) — all are Runic style normalizations that preserve behavior (broadcast dot retained, `for x = r`/`for x in r` are equivalent Julia syntax).
- `docs/make.jl` gained an explicit `return` at the end of `clear_md_files`; the function already implicitly returned `nothing` there, so this is a no-op.

No blocking issues.

---

src/util/truncated_normal.jl
cossio: Yes. Updated in `28802bed`: `randnt` now keeps the outer `if` and ends each branch with its own `return r`. I rewrote each rejection loop as an initial draw followed by the same retry predicate, so the random draw order and acceptance conditions are unchanged (including the negated comparisons used for NaN behavior). Pinned Runic 1.7 accepts this form unchanged. Validation: `test/truncnorm.jl` passed 32/32, the full package suite passed, and Runic check mode passed all tracked Julia files.

src/util/truncated_normal.jl
```diff
@@ -20,20 +20,21 @@ randnt(a::Real) = randnt(default_rng(), a)
...
+        while !(u < exp(-(r - t)^2 / 2))
             r = a + randexp(rng, typeof(a)) / t
             u = rand(rng, typeof(a))
-            if u < exp(-(r - t)^2 / 2)
-                return r
-            end
         end
+        return r
     end
 end
```
cossio, line 37: It seems to me that the original version of `randnt` passes Runic, so I'd prefer that. It had much less duplication

cossio: Agreed, reverted in `d577af2`: `randnt` is back to the mechanical Runic output (`return if ... while true ... end`), which is behaviorally identical and avoids the duplicated initial draw. `test/truncnorm.jl` and the Runic check both cover this form already from the earlier `bff14c9` commit.

---

cossio: Follow-up correction in `576fa478`: the exact pre-PR file does not pass pinned Runic 1.7 (`git show e77b1b24:src/util/truncated_normal.jl | julia --project=@runic --startup-file=no -m Runic --check --diff -` exits 1 and changes the outer `if` to `return if`). I applied the no-duplication structure suggested in the inline comments instead: one typed local `r`, `while true` in both branches, `break` on acceptance, then branch-local `return r`. The typed local is required because a variable first assigned inside a Julia loop is otherwise unavailable after the loop. Runic check mode passes, and `test/truncnorm.jl` passes all 32 assertions including `@inferred` and edge cases.

cossio (commented):
## Review

Checked the actual commit history (`get_commits`) against the PR description's claims, and ran a whitespace-insensitive diff over the full patch to isolate any hunk where non-whitespace tokens (not just spacing) changed.

### ⚠️ Blocking: PR description is stale — this is no longer a pure mechanical-formatting change

`get_commits` shows 5 commits, not the 2 described:
1. `8044b179` — Apply Runic formatting (07:57)
2. `bff14c968` — Enforce Runic formatting in CI (07:58)
3. **`28802bed`** — "Use branch-local returns in randnt" (09:17)
4. **`d577af29`** — "Revert randnt..." (09:22)
5. **`576fa478`** (current HEAD) — "Avoid duplicated randnt draws" (09:24)

The PR was opened at 08:00:15Z, so the last three commits landed *after* the description/validation section was written, and the description was never updated to reflect them. I fetched `src/util/truncated_normal.jl` at each SHA:
- `8044b179` (the claimed "isolated mechanical commit") produces the plain Runic-reformatted control flow.
- `28802bed` hand-rewrites this to a hoisted-draw-before-loop pattern — and in doing so **introduces a real bug** (the initial draw ends up computed/discarded twice).
- `d577af29` reverts back to the exact Runic output.
- `576fa478` (HEAD) hand-rewrites it again (`local r::typeof(a); ... r ≥ a && break; end; return r`) to fix the duplicated-draw issue without reintroducing it.

So there's a genuine, deliberate, non-mechanical source edit — with a bug introduced and fixed within the same branch — sitting inside a PR whose entire premise is "pure formatting, safe to rubber-stamp." The final state looks behaviorally equivalent to the pre-PR baseline, but this should either be disclosed in the PR body or split out into its own PR so `randnt`'s control-flow rewrite gets reviewed on its own merits.

**Suggestion:** No regression test was added asserting draw-count/determinism for `randnt` under a seeded RNG, even though this exact function had a duplicated-draw bug introduced and fixed mid-branch. `test/truncnorm.jl`'s diff is 100% numeric-literal reformatting (`1e-6` → `1.0e-6` etc.) — worth adding a targeted test for the fixed behavior.

### Verified OK
- `.git-blame-ignore-revs` correctly lists only `8044b179b339b41ca526ee24520c8bedf4941ece` (the true mechanical commit), not the CI or randnt commits.
- `.github/workflows/format.yml`: `actions/checkout`, `setup-julia`, and `fredrikekre/runic-action` are all pinned by full 40-char SHA (with `# v1.4.0` as a human-readable comment). Triggers on `push`/`pull_request` with no path filters, so future new files are covered automatically. (Could not independently verify the runic-action SHA against upstream `v1.4.0` — this session's repo access is scoped to this repository only.)
- Spot-checked reformatting across layers, training code, Literate docs sources, and tests: resolves entirely to known-safe Runic transforms (spacing, numeric literal completion, `for x = it` → `for x in it`, explicit `return` insertion, line-reflow) — no operator-precedence changes or comment relocations altering meaning found outside the `randnt` case above.
- `#hide` Literate directives preserved verbatim in `docs/src/literate/{MNIST,ais,metropolis}.jl` and layer docs.
- No `CHANGELOG.md` entry added — correct, per repo convention.

### Verdict

The reformatting itself is clean and verified safe. I'd hold off on merging as-is on process grounds: please update the description to disclose the `randnt` rewrite (or split it into its own PR), and consider adding a regression test for the bug that was introduced and fixed mid-branch. Also worth confirming the full test matrix is green on the current head SHA (`576fa478`) before merging, given how late those commits landed relative to the stated validation section.

---

cossio (commented):
Agree this is a valid blocking concern: the description still reads as a 2-commit, pure-mechanical change (5 commits now on HEAD `576fa478`, and `randnt` did get a genuine hand-written control-flow rewrite mid-branch, including a bug introduced then fixed). The description hasn't been updated yet to disclose this.

Two more things worth flagging before merge:
- `mergeable_state` is now `dirty` — this branch conflicts with `master` (likely because #164 just merged and touched overlapping files/history). Will need a rebase/merge of `master` before this can land.
- No regression test pins down `randnt`'s draw count/determinism under a seeded RNG, despite a duplicated-draw bug having been introduced and fixed on this same branch. A targeted `@testset` with a fixed `MersenneTwister`/`Xoshiro` seed asserting the exact sequence of values (or number of `rand`/`randn` calls consumed) would catch a regression of that specific bug if Runic or a future edit touches this function again.

CI (Runic, CodeQL) is green on `576fa478`; the Julia test matrix was still running at last check.

---

cossio (commented):
## Review

**What it does:** Applies Runic 1.7 formatting to all 86 tracked `.jl` files in one isolated mechanical commit, adds a CI check via `fredrikekre/runic-action` (pinned to a commit SHA), and records the formatting commit in `.git-blame-ignore-revs` so `git blame` stays useful.

**Checked:**
- Spot-checked several large source diffs (`src/layers/drelu.jl`, `src/util/util.jl`, `docs/make.jl`) — all changes are pure whitespace/style normalization (spacing around `<:`, `/`, commas in type params; `1e300` → `1.0e300`; `for a = ...` → `for a in ...`), nothing semantic.
- `.github/workflows/format.yml` pins both the checkout/setup-julia actions and `runic-action` to commit SHAs (with the `v1.4.0` tag as a comment) — good supply-chain hygiene, follows the version-pinning pattern already used elsewhere in the repo.
- The commit history shows some non-mechanical back-and-forth on `src/util/truncated_normal.jl`'s `randnt` (a branch-local-return rewrite was tried, then reverted in favor of the original Runic-hoisted-return form, then a final "avoid duplicated draws" tweak). I compared the final diff against the pre-PR version by hand: the refactor from early-return-inside-`while true` to `local r::typeof(a); ...; break; ...; return r` is behavior-preserving — same draw/accept-reject logic, no duplicated `randn`/`randexp` draws, same fallback paths for `a ≤ 0` vs the `else` branch.
- No CHANGELOG entry, correctly per repo convention (CI/tooling-only change, no user-facing behavior change).

**Minor note (non-blocking):** the PR description says "The follow-up commit contains only CI/tooling metadata," but the branch has 5 commits, including 3 that touch `randnt`'s control flow after the initial formatting+CI commits. Worth a one-line addendum to the description for future readers, but the code itself is correct and well-tested (full suite, `@inferred` checks on `randnt` pass per the PR's own validation notes).

No issues found. LGTM (posting as comment — GitHub blocks self-approval on your own PR).

---

cossio (commented):
No blocking issues. I diffed `origin/master` against this branch with `-w --ignore-blank-lines` to separate real edits from Runic's whitespace/line-wrapping output, and everything in `src/`, `ext/`, and `test/` reduces to formatting (numeric-literal normalization like `1e-6`→`1.0e-6`, `for x = y`→`for x in y`, added trailing `return`, line wrapping) with one exception.

**Nit:** `src/util/truncated_normal.jl`'s `randnt(rng, a::Base.IEEEFloat)` was manually restructured in the follow-up commits (`28802be`, `d577af2`, `576fa47`, `c3d8f02`, `bcc68fa` — titled "Use branch-local returns", "Avoid duplicated randnt draws", "Simplify randnt return flow/non-finite handling"), not produced by Runic. This contradicts the PR description's claim that "the follow-up commit contains only CI/tooling metadata" and that "no manual rewrites... were needed." The rewrite drops the explicit `!(t < Inf) && return a` early-exit and dispatches on `elseif a < Inf` instead. I checked this for a behavior change: `sqrt1half`/`_sqrt1half` (same file) only takes the `x > 2/sqrt(eps(x))` branch and returns `x` itself, which is finite whenever `x`/`a` is finite — so `t = sqrt1half(a)` can never actually be `Inf` for finite `a`, and the removed check was already dead code. The `a == Inf` and `a == NaN` cases are still handled correctly via the new `local r = a` initialization. So the rewrite looks behavior-preserving, and `test/truncnorm.jl` already exercises the relevant edge cases (`randnt(Inf)`, `randnt(NaN)`, `randnt(floatmax(Float64))`, `randnt(1e300)`), which CI covers. Still, worth correcting the PR description so reviewers don't assume the diff is 100% mechanical formatter output.

---

cossio:
Addressed the latest automated review findings. The PR description now discloses the deliberate post-formatting `randnt` rewrite, current `master` integration, deterministic regression coverage, and updated validation instead of describing the branch as only two pure-mechanical commits. The conflict with `master` was resolved in `9379d39e` while preserving the exact formatting commit `8044b179b339b41ca526ee24520c8bedf4941ece`. The final `randnt` form is in `bcc68fa7`, and `9e149025` adds a deterministic test RNG that asserts the exact accepted/rejected draw sequence and verifies that `Inf`/`NaN` consume no draws. Locally: all 86 tracked Julia files pass pinned Runic 1.7 check mode; `test/truncnorm.jl` passes 39/39; affected pseudolikelihood and JLArrays tests pass with only the pre-existing broken Potts test; the full package suite passes with no failures.

src/util/truncated_normal.jl
cossio: Final resolution is now on the PR: `randnt` uses the requested `local r = a`, two `while true`/`break` rejection loops, `elseif a < Inf`, and one final `return r` (`bcc68fa7`). This avoids duplicated draws; `Inf` and `NaN` skip both sampler bodies and return the initialized `r` without consuming RNG state. Pinned Runic 1.7 accepts the code unchanged. Commit `9e149025` adds a deterministic test RNG that verifies the exact negative- and positive-branch draw sequences plus zero draws for non-finite inputs. Validation: focused truncated-normal suite 39/39, repository-wide Runic check clean, and full package suite passed with the one pre-existing broken Potts test and no failures.

cossio (commented):
## Re-review after latest push (`9e14902`)

Independently re-verified the two things worth checking on this update:

1. **`randnt` non-finite handling (`bcc68fa`)**: confirmed `sqrt1half`/`_sqrt1half` can never return `Inf` for finite input — `_sqrt1half(x)` either returns `x` itself (when `x` is above the large-`x` threshold, so still finite) or `sqrt(1+(x/2)^2)+x/2` (bounded for any finite `x`, no overflow before the threshold kicks in). So the old `!(t < Inf) && return a` guard was dead code for finite `a`, and dropping it in favor of `local r = a` + `elseif a < Inf` is behavior-preserving. The new deterministic-RNG test (`9e14902`, `TruncatedNormalTestRNG`) directly confirms both rejection-loop draw sequences and that `Inf`/`NaN` consume zero RNG draws — good regression coverage for exactly the thing that changed.
2. **Merge with `master` (`9379d39`, resolving the `src/pseudolikelihood.jl` conflict against PR #164's Gaussian pseudolikelihood work)**: diffed `origin/master` against this branch with `-w --ignore-blank-lines` and confirmed the entire `pseudolikelihood.jl` delta is Runic reformatting (indentation, `dims=1` → `dims = 1`, added `return`) — no conflict markers left behind, no logic changes.

PR description now accurately reflects the hand-rewrite and merge, addressing the earlier nit. No blocking issues from me either. CI is still finishing on the latest push but Runic checks are already green.

---

PR #166, opened 2026-07-19
Enforce explicit imports in CI

cossio:
## Summary

- add ExplicitImports.jl 1.15 as a test-only dependency and run its full stable check set from the existing test suite
- check the package module plus the HDF5 and CUDA extensions; exercise the actual CUDAExt source with a lightweight `CUDA` stub carrying a non-CUDA UUID in an isolated subprocess, so CI needs neither CUDA.jl nor a physical GPU and unrelated CUDA extensions cannot activate
- fix the namespace findings mechanically: remove stale imports, replace private/self-qualified names, and remove private `HDF5.File` annotations from internal loader helpers

## Deliberate compatibility and exclusions

The `ExplicitImports = "1.15"` compat floor is deliberate because `test_explicit_imports`, which aggregates the current seven stable checks, was added in ExplicitImports 1.15.

The checks stay strict except where package architecture or upstream metadata makes a public-name assertion invalid:

- on Julia 1.10 only, ignore `Random.default_rng` and `Optimisers.setup`/`update!` for the public-import check because Julia 1.10 cannot represent `public` bindings; the ignore is empty on Julia 1.11+
- ignore Adapt's documented `@adapt_structure` integration macro for the public-qualified-access check because Adapt does not mark it public
- scope extension public-name checks to their third-party dependencies (`HDF5`, and `CUDA`/`Adapt`); RestrictedBoltzmannMachines intentionally exports no symbols, while its extensions necessarily import and extend package internals

All ownership, implicit-import, stale-import, and self-qualified-access checks remain enabled for the package and both extensions.

## Validation

- `julia --project=test test/explicit_imports.jl`
- `julia +1.10 --project=test test/explicit_imports.jl`
- exact downgrade-compat manifest reproduction followed by the full focused ExplicitImports gate (16/16)
- focused touched paths: `test/onehot.jl`, `test/layers/nsReLU.jl`, `test/truncnorm.jl`, `test/hdf5.jl`, and `test/train.jl`
- integration probe with NNlib and Zygote preloaded, confirming the CUDA fixture stays subprocess-isolated
- `julia --project=. -e 'using Pkg; Pkg.test()'`
- `git diff --check`

No changelog entry: this is a CI/test enforcement change with behavior-preserving mechanical source fixes.


src/util/truncated_normal.jl
```diff
@@ -18,7 +18,7 @@ randnt(rng::AbstractRNG, a::Real) = randnt(rng, float(a))
 randnt(rng::AbstractRNG, a::BigFloat) = randnt(rng, Float64(a))
 randnt(a::Real) = randnt(default_rng(), a)
 
-function randnt(rng::AbstractRNG, a::Base.IEEEFloat)
+function randnt(rng::AbstractRNG, a::Union{Float16,Float32,Float64})
```
cossio: Nit (non-blocking): replacing `Base.IEEEFloat` with the inlined `Union{Float16,Float32,Float64}` is behaviorally identical today, but it silently hardcodes an implementation detail of `Base` that this file no longer references by name. If `Base.IEEEFloat`'s definition ever changes, this won't track it. Consider a one-line comment noting it mirrors `Base.IEEEFloat` (which ExplicitImports flags as a non-public qualified access), so a future reader knows why the union is spelled out instead of just widening the dispatch.

---

ext/HDF5Ext.jl
cossio, line 1: Verified this is safe: `_load_rbm` is only ever invoked from `load_rbm` (line ~34) with the `file` object handed in by `h5open(path, "r") do file ... end`, so `file` is always an `HDF5.File` at the only call site regardless of the annotation. The three `_load_rbm` methods are disambiguated by the `::Val{...}` second argument, not by the type of `file`, so dropping the `HDF5.File` annotation doesn't introduce dispatch ambiguity or change which method runs for a given saved-format tag. No loss of the malformed-file runtime check — `read(file, "weights")` etc. would already fail the same way for a non-HDF5.File argument with or without the annotation. Looks fine as mechanical cleanup.

---

cossio (commented):
Reviewed per REVIEW.md priorities: correctness/behavior-preservation of the "mechanical" import fixes, HDF5 compatibility risk, and soundness of the CUDA extension test isolation. No blocking issues found.

**HDF5Ext.jl (compatibility-sensitive):** Removing the `HDF5.File` annotations from the three internal `_load_rbm` overloads is safe. `_load_rbm` is only called at one site (`load_rbm`, inside the `h5open(...) do file` block), so `file` is always an `HDF5.File` regardless of the annotation, and the three methods are disambiguated by `::Val{...}`, not by the type of `file`. No dispatch ambiguity introduced, no runtime malformed-file check lost — `read(file, ...)` fails the same way either way. See inline note.

**Stale imports:** Spot-checked every removed symbol (`front`, `tail`, `Fill`, `Trues`, `norm`, `std`, self-qualified `RestrictedBoltzmannMachines.AbstractLayer`) against `src/` and `ext/` with `git grep`; none are used anywhere outside the import lines that were removed. The one `std(...)` mention that remains (in `src/ais.jl`'s docstring) is plain prose, not a `jldoctest`/`@example` block, so it doesn't execute and doesn't need the import. The `tail(size(x))` → `size(x)[2:end]` rewrites are behaviorally equivalent (tuple range indexing is type-stable/constant-folded here) and `Base.IEEEFloat` → `Union{Float16,Float32,Float64}` matches its current definition exactly (minor nit posted inline).

**CUDA extension test isolation:** Confirmed sound, not a false-pass risk. The fixture's `CUDA` stub uses UUID `4a9ad4d1-...`, distinct from the real CUDA.jl UUID (`052768ef-...`) declared in the root `Project.toml`'s `[weakdeps]`, so Julia's extension-triggering mechanism (UUID-based) cannot activate the real `CUDAExt` — verified by the test's own `@test isnothing(Base.get_extension(RBMs, :CUDAExt))` assertion. Crucially, `include(extension_file)` loads the actual `ext/CUDAExt.jl` source text (not a mock), so genuine future changes to that file are exercised by the explicit-imports check. Subprocess isolation is appropriately used only to keep the stub's `LOAD_PATH` mutation from leaking into the main test process.

**Ignore-lists:** Both are narrowly scoped and justified — the Julia-1.10-only `default_rng`/`setup`/`update!` ignore is conditioned on `VERSION < v"1.11"` and empties out on newer Julia where `public` bindings are representable; the `@adapt_structure` ignore is a single documented-but-not-`public` macro from Adapt. Neither is a blanket suppression.

Two non-blocking nits posted inline (well under the cap); nothing here rises to a blocking concern.

---

src/util/truncated_normal.jl
cossio: Addressed in 43b0631c: added a one-line comment documenting that the explicit Float16/Float32/Float64 union mirrors private Base.IEEEFloat and is spelled out to avoid relying on a non-public name. The focused truncnorm and ExplicitImports tests both pass.

ext/HDF5Ext.jl
cossio: Thanks for verifying this. Agreed: the Val argument remains the method discriminator, the only caller still supplies HDF5.File, and read failures are unchanged. No additional code change is needed; the focused HDF5 coverage and full suite pass.

cossio (commented):
Reviewed the ExplicitImports enforcement and its mechanical fixes.

- `randnt(rng, a::Base.IEEEFloat)` → `Union{Float16,Float32,Float64}`: this is an exact, non-behavior-changing substitution (`Base.IEEEFloat` is defined as exactly this union), with a comment explaining why.
- `tail(size(x))` → `size(x)[2:end]` (gradient.jl, onehot.jl): equivalent for tuples.
- `_load_rbm(file::HDF5.File, ...)` → `_load_rbm(file, ...)`: these are internal (`_`-prefixed) helpers only ever called with actual `HDF5.File`/group values at their call sites, so loosening the annotation doesn't change dispatch behavior or touch the on-disk HDF5 schema.
- Removed unused imports (`Fill`, `Trues`, `norm`, `std`, `front`) — full test suite passing confirms nothing still referenced them.
- The CUDA ExplicitImports fixture runs in a genuinely separate subprocess (`test/explicit_imports_cuda.jl` invoked via `run(...)`), so the stub `CUDA` module and its `LOAD_PATH` mutation never touch the main test process, and no physical GPU or real CUDA.jl is required — consistent with the repo's CI constraints.
- Public-name-check scoping (ignoring `default_rng`/`setup`/`update!` only on Julia < 1.11, scoping extension checks to `HDF5`/`CUDA`/`Adapt` since the package itself exports nothing) matches the package's documented "exports no symbols" architecture.

No correctness issues found. No blocking issues.

---

cossio (commented):
## Review

Verified the "mechanical" fixes against actual usage sites and CI results rather than taking the description at face value.

**Verified safe:**
- Removed imports (`tail`, `front`, `Fill`, `Trues`, `norm`, `std`) in `src/RestrictedBoltzmannMachines.jl`: grepped `src/` for remaining usages — none found (the one `std(` hit is inside a docstring, not code).
- Dropped `file::HDF5.File` annotations on `_load_rbm` in `ext/HDF5Ext.jl`: the three methods dispatch on the `Val{...}` second argument, not `file`'s type, so removing the annotation doesn't change dispatch. Minor loss of self-documentation/early type-error at the boundary, but not a real risk (already raised and addressed in a resolved thread).
- `Base.IEEEFloat` → `Union{Float16,Float32,Float64}` in `src/util/truncated_normal.jl`: semantically identical today.
- `tail(size(x))` → `size(x)[2:end]`, and dropping same-module self-qualification (`RestrictedBoltzmannMachines.AbstractLayer{N}` → `AbstractLayer{N}`): no behavior change.
- CUDA stub isolation (`test/explicit_imports_cuda.jl`, `test/fixtures/CUDA/*`): the stub UUID (`4a9ad4d1-...`) is confirmed distinct from the real CUDA.jl UUID (`052768ef-...`) in the root `Project.toml`'s `[weakdeps]`, so it cannot activate the real `CUDAExt` via Julia's UUID-keyed extension mechanism. Subprocess isolation correctly prevents the stubbed `LOAD_PATH` from leaking into the parent test process.
- `ExplicitImports.jl` API usage (`test_explicit_imports`, `ignore` keyword) checked against upstream source — not fabricated.
- `test/explicit_imports.jl` is wired into `test/runtests.jl`, so it runs under standard `Pkg.test()`, matching the PR description.
- No `CHANGELOG.md` entry added, consistent with the repo convention that CI/tooling changes don't need one.
- CI is green on the head SHA (`43b0631`) across `Julia 1.10`, `Julia 1`, and `Downgrade compat`.

**Nitpicks (non-blocking):**
- The ignore-list scope is narrow and justified (3 symbols gated to Julia <1.11 only, one Adapt macro not marked public upstream) — no overly broad exclusions that could mask future real findings.
- `test/explicit_imports_cuda.jl`'s `try/finally` cleanup around `popfirst!(LOAD_PATH)` is unnecessary ceremony since the subprocess exits immediately after — harmless.

No security concerns: subprocess invocation uses `Base.julia_cmd()` with no untrusted input interpolation, and the fixture UUID is never resolved through Pkg's registry.

**Verdict: looks good to merge** (note: I can't formally "Approve" via this token since it's authenticated as the PR author; treat this as an approving review in substance).

---

cossio (commented):
## Review

**What it does:** Adds `ExplicitImports.jl` as a test-only dependency, wires its full stable check set into `test/runtests.jl`, and fixes the handful of namespace issues it surfaces (stale imports, self-qualified names, a private-type annotation).

**Checked against the local checkout (which matches this PR's head, `0b4bf1b`):**
- `src/RestrictedBoltzmannMachines.jl`: removed `Base: front, tail`, `FillArrays: Fill, Trues`, `Statistics: std`. Grepped `src/` — none of `Trues`, `Fill(`, `tail(`, `front(`, or a code (non-docstring) use of `std(` remain, so these removals are safe (no latent `UndefVarError`).
- `src/util/onehot.jl` (`categorical_sample`) and `src/train/gradient.jl` (`∂RBM` constructor): `Base.tail(size(x))` → `size(x)[2:end]`. `categorical_sample` is a sampling hot path, so I checked type-stability empirically on Julia 1.12.6 — both forms infer to the same concrete `Tuple{Int,...}` type (`Base.return_types` and `@inferred` agree). No instability introduced.
- `ext/HDF5Ext.jl`: `_load_rbm(file::HDF5.File, ...)` → `_load_rbm(file, ...)`, dropping the type annotation because `HDF5.File` isn't part of HDF5's public API surface per the check's metadata. This is a private internal helper only ever called with the `file` handle from `h5open` in the same module, so the loosened dispatch is low-risk. A slightly tighter alternative would have been an explicit `ignore` entry for `:File` in the `all_qualified_accesses_are_public` config (keeping the annotation), but the chosen fix is reasonable and doesn't affect correctness.
- `src/util/truncated_normal.jl`: `Base.IEEEFloat` → `Union{Float16,Float32,Float64}`, which is exactly `Base.IEEEFloat`'s definition, with a comment explaining why. Exact behavioral equivalent.
- `src/layers/nsReLU.jl`: dropped self-qualification (`RestrictedBoltzmannMachines.AbstractLayer{N}` → `AbstractLayer{N}`) — correct, purely cosmetic.
- `test/explicit_imports_cuda.jl`: nice isolation approach — a non-CUDA-UUID stub package pushed onto `LOAD_PATH`, run in a genuinely separate `julia` subprocess (via `run(... ignorestatus(...))`), so it can exercise the real `ext/CUDAExt.jl` source without needing CUDA.jl or GPU hardware, and without leaking `LOAD_PATH` state into the main test process.
- The `all_qualified_accesses_are_public` ignore for `Symbol("@adapt_structure")` is justified — confirmed `Adapt.@adapt_structure` is actually used in `src/adapt.jl`.
- No CHANGELOG entry, correctly per repo convention (test/CI enforcement only, no source behavior change beyond the mechanical import cleanup).

No issues found. LGTM (posting as comment — GitHub blocks self-approval on your own PR).

---

cossio (commented):
No blocking issues.

Reviewed the mechanical fixes against REVIEW.md's correctness bar:
- `src/RestrictedBoltzmannMachines.jl`: removed imports (`front`, `tail`, `Fill`, `dot`... wait, kept `dot`) match what's actually used elsewhere — spot-checked `tail`/`front` are no longer referenced in `src/`.
- `∂RBM` constructor (`src/train/gradient.jl`) and `categorical_sample` (`src/util/onehot.jl`): `tail(size(x))` → `size(x)[2:end]` is equivalent for the array shapes these functions actually receive (always ≥1 dimension in this codebase), so this isn't a behavior change in practice.
- `randnt(rng, a::Base.IEEEFloat)` → `Union{Float16,Float32,Float64}`: this union is exactly `Base.IEEEFloat`'s definition, so dispatch is unchanged.
- `_load_rbm(file::HDF5.File, ...)` → `_load_rbm(file, ...)`: only called internally with `HDF5.File` arguments, so widening the signature doesn't change behavior.
- The new `test/explicit_imports.jl`/`test/explicit_imports_cuda.jl` harness correctly isolates the CUDA stub in a subprocess with a non-CUDA UUID, avoiding interference with a real CUDA.jl extension.

No changelog entry is appropriate here per CLAUDE.md (CI/test enforcement, behavior-preserving).

---

ext/HDF5Ext.jl
```diff
@@ -122,14 +122,14 @@ construct_layer(::Val{:xReLU}, par::AbstractArray) = xReLU(par)
 construct_layer(::Val{:nsReLU}, par::AbstractArray) = nsReLU(par)
 construct_layer(::Val{:PottsGumbel}, par::AbstractArray) = PottsGumbel(par)
 
-function _load_rbm(file::HDF5.File, ::Val{:RBM})
+function _load_rbm(file, ::Val{:RBM})
```
cossio, line 125: why are you removing the `file::HDF5.File` type annotation here?

src/train/gradient.jl
```diff
@@ -3,7 +3,7 @@ struct ∂RBM{V,H,W}
     hidden::H
     w::W
     function ∂RBM(visible::AbstractArray, hidden::AbstractArray, w::AbstractArray)
-        @assert size(w) == (tail(size(visible))..., tail(size(hidden))...)
+        @assert size(w) == (size(visible)[2:end]..., size(hidden)[2:end]...)
```
cossio, line 6: what's wrong with using `tail` ? Is not considered stable API?

ext/HDF5Ext.jl
cossio: `HDF5.File` isn't marked public in HDF5.jl's metadata, so `ExplicitImports.test_explicit_imports`'s `all_qualified_accesses_are_public` check flags annotating with it by name as a non-public qualified access — that's what this mechanical pass is fixing.

It's safe to drop here: `_load_rbm` is a private (`_`-prefixed) helper called from exactly one place, `load_rbm`, inside `h5open(path, "r") do file ... end` (so `file` is always an `HDF5.File` at the call site regardless of the annotation). The three `_load_rbm` methods are disambiguated by the `::Val{:RBM}` / `::Val{:StandardizedRBM}` / `::Val{:CenteredRBM}` second argument, not by the type of `file`, so removing the annotation introduces no dispatch ambiguity and doesn't change which method runs. `read(file, ...)` would fail the same way for a malformed argument with or without it.

This was also raised on the sibling thread on this file (verified there in more detail) — happy to add an `ignore` entry for `:File` in the `all_qualified_accesses_are_public` config instead if you'd rather keep the explicit annotation; that's the more surgical alternative to loosening the signature.

---

src/train/gradient.jl
cossio: Right — it's not that `tail` is unstable in the sense of "likely to change soon," it's that `Base.tail`/`Base.front` are neither exported nor marked `public` (Julia 1.11+'s explicit `public` keyword), so they carry no semver/API-stability guarantee at all, even though they've been around a long time. `ExplicitImports.jl`'s `all_explicit_imports_are_public` check flags `using Base: front, tail` for exactly that reason: it's importing internal Base implementation details by name.

`tail(t)`/`front(t)` are just `t[2:end]`/`t[1:end-1]` for a `Tuple`, so this PR replaces the two call sites (here in `∂RBM`'s constructor, and in `categorical_sample` in `src/util/onehot.jl`) with `size(...)[2:end]` and drops the import — same result, public API only. Tuple range-indexing is type-stable/compiles away for the fixed-size tuples these functions see, so there's no performance or correctness change, just avoiding a dependency on an unexported Base internal.

---

test/fixtures/CUDA/src/CUDA.jl
cossio, line 1: what is the purpose of this file?

cossio: It's a lightweight stand-in `CUDA` package used only to let `test/explicit_imports_cuda.jl` load and lint `ext/CUDAExt.jl` in CI without needing the real CUDA.jl or physical GPU hardware.

`ext/CUDAExt.jl` is a package extension that only activates when both `CUDA` and `Adapt` are loadable, so ExplicitImports can't check it unless something named `CUDA` is present. This fixture (`test/fixtures/CUDA/`) is a minimal package with a `Project.toml` declaring `uuid = "4a9ad4d1-..."` — deliberately *not* the real CUDA.jl UUID (`052768ef-...`) — plus this one-line `src/CUDA.jl` exporting a no-op `cu`. Because Julia's extension-triggering is keyed on UUID, this stub can never accidentally activate the real `CUDAExt`; `test/explicit_imports_cuda.jl` asserts that (`@test isnothing(Base.get_extension(RBMs, :CUDAExt))`) and instead `include()`s the actual `ext/CUDAExt.jl` source directly so its real imports get linted.

It's pushed onto `LOAD_PATH` and run in a separate `julia` subprocess (`test/explicit_imports_cuda.jl` invoked via `run(...)` from `test/explicit_imports.jl`) so the `LOAD_PATH` mutation and stub module never leak into the main test process.

---

cossio (commented):
Reviewed the diff and checked out `codex/enforce-explicit-imports` locally to verify the mechanical fixes.

**Verified safe:**
- Removed imports (`Base.front`/`tail`, `FillArrays.Fill`/`Trues`, `LinearAlgebra.norm`, `Statistics.std`) have no remaining call sites in `src/` or `ext/` (grepped to confirm).
- `tail(size(x))` → `size(x)[2:end]` in `gradient.jl`/`onehot.jl` is behavior-preserving and stays type-stable for tuples.
- `Base.IEEEFloat` → `Union{Float16,Float32,Float64}` in `truncated_normal.jl` is exactly equivalent (that's literally `Base.IEEEFloat`'s definition).
- Dropping the `::HDF5.File` annotations on the three internal `_load_rbm` methods is safe: they're only ever called from the `h5open(...) do file` closure in the same file, and each `Val` type has a single dispatch target, so genericizing the argument doesn't introduce ambiguity.
- The CUDA fixture (`test/fixtures/CUDA`) uses UUID `4a9ad4d1-...`, distinct from the real CUDA.jl UUID (`052768ef-...`), so it can't be mistaken for the real package, and running it via subprocess with an isolated `LOAD_PATH` keeps it from leaking into the rest of the suite.
- `nsReLU.jl`'s self-qualification removal (`RestrictedBoltzmannMachines.AbstractLayer{N}` → `AbstractLayer{N}`) is a no-op inside the module.

**Process notes:**
- No `CHANGELOG.md` entry, consistent with repo policy (CI/test-only change, no public API or behavior change).
- CI was still `in_progress` on the head commit (`2c1c750`, "Format ExplicitImports changes with Runic") at review time — worth confirming green before merge, but I don't see anything in the diff itself that would fail it.

No issues found. (Can't formally approve — this account matches the PR author — but this reads as ready to merge once CI is green.)

---

Issue #170, opened 2026-07-22
Refactor: collapse RBM/CenteredRBM/StandardizedRBM duplication (~300–350 lines)


cossio:
Full-repo review surfaced heavy structural duplication across `RBM`, `CenteredRBM`, and `StandardizedRBM` (`src/rbm.jl` 399 lines, `src/centered.jl` 531, `src/standardized.jl` 608, plus `src/train/`). Estimated realistic reduction: **~300–350 lines**, in three escalating steps.

Key architectural fact: the generic methods in `src/rbm.jl` are untyped in `rbm` (e.g. `sample_h_from_v(rbm, v)`), and all polymorphism flows through dispatch on `inputs_h_from_v` / `inputs_v_from_h` / `interaction_energy`. `StandardizedRBM` already exploits this (it defines no `sample_*`, `mean_*`, or `energy(rbm, v, h)` methods); `CenteredRBM` does not, and re-copies them.

## Step 1 — delete verbatim duplicates (~90 lines, near-zero risk)

These `CenteredRBM` methods are textually identical to the untyped generics in `src/rbm.jl` and would dispatch identically if deleted:

| Method | centered.jl | duplicate of rbm.jl |
|---|---|---|
| `energy(rbm, v, h)` | 82–87 | 89–94 |
| `mean_h_from_v`, `mean_v_from_h` | 113–121 | 198–211 |
| `sample_h_from_v`, `sample_v_from_h` | 406–414 | 138–151 |
| `sample_v_from_v`, `sample_h_from_h` | 416–428 | 159–179 |
| `sample_v_from_v_once`, `sample_h_from_h_once` | 430–440 | 181–191 |

Proof of safety: `StandardizedRBM` relies on exactly these generics and is fully covered by `test/standardized.jl`. (Minor behavior delta: the generic `sample_v_from_v`/`sample_h_from_h` add a size `@assert` the centered copies lack — strictly an improvement.)

Also in this step:

[...]

PR #175, opened 2026-07-22
Collapse RBM/CenteredRBM/StandardizedRBM method duplication (#170, steps 1–2)

cossio:
Implements **steps 1 and 2** of #170. Net effect: −422 / +268 lines, with the duplicated `CenteredRBM`/`StandardizedRBM` method bodies collapsed into untyped generics and a shared `Union{CenteredRBM, StandardizedRBM}` implementation file.

## Step 1 — delete verbatim duplicates

- Deleted the `CenteredRBM` methods that were textually identical to the untyped generics in `src/rbm.jl`: `energy(rbm, v, h)`, `mean_h_from_v`/`mean_v_from_h`, and the six `sample_*` methods. Dispatch now flows through the generics, exactly as `StandardizedRBM` already did.
- Untyped `∂free_energy`, `∂free_energy_v`, and `∂free_energy_h` in `src/train/gradient.jl` and deleted the character-identical copies in `centered.jl` and `standardized.jl`.
- `mirror(::CenteredRBM)` now delegates through `mirror(RBM(rbm))` instead of re-copying the `perm` computation.
- `rescale_weights!` had three identical bodies (plain/centered/standardized); one untyped method remains.

## Step 2 — share Centered/Standardized code via unit-scale accessors

New `src/offset_rbms.jl` defines `const OffsetRBM = Union{CenteredRBM, StandardizedRBM}` and writes each of these once: `interaction_energy`, `inputs_h_from_v`/`inputs_v_from_h`, `free_energy`, `free_energy_h`, `∂interaction_energy`, `log_pseudolikelihood`, `∂regularize!` (plus the `∂regularize_add_visible_offset!` helpers), `regularization_penalty`, `zerosum!(rbm)`, and `zerosum!(∂, rbm)`.

`CenteredRBM` exposes its unit scales as `FillArrays.Ones` via internal `_scale_v`/`_scale_h`/`_scale_w` accessors, and `_maybe_div`/`_maybe_mul` specializations skip the scaling no-ops (verified: `reshape(::Ones)` and the accessors stay `Ones`), so the centered hot paths perform no divisions by one and no extra allocations relative to the deleted hand-written versions. Plain `RBM` methods are untouched.

## Behavior changes (all user-facing, in CHANGELOG)

- **Fix**: `free_energy_h(::CenteredRBM, h)` previously fell through to the plain `RBM` generic, omitting the `energy(Binary(θ = offset_v), inputs)` correction that `StandardizedRBM` includes. It now includes it, making `free_energy_h` consistent with `free_energy(mirror(rbm), h)`, with exact marginalization over visible states, and with the equivalent uncentered RBM up to the constant gauge shift — all covered by new tests.
- `∂free_energy_v`, `∂free_energy_h`, and `regularization_penalty` now accept `CenteredRBM` (previously `MethodError`); `∂free_energy_h(::CenteredRBM)` is verified against Zygote gradients of the (fixed) `free_energy_h`.
- `∂regularize!(∂, ::CenteredRBM)` gains the `regularize_unstandardized` keyword (default `true` preserves the old behavior).
- `sample_v_from_v`/`sample_h_from_h` for `CenteredRBM` now assert configuration sizes via the generics (the issue notes this as strictly an improvement).

`Project.toml` is bumped to `5.9.0-DEV` as a development placeholder (new capabilities for `CenteredRBM` suggest a minor bump) — final version is of course the owner's call at release time.

## Step 3 not included

The training-loop skeleton unification (`pcd!` ×3 + `ucd!`) is intentionally left for a follow-up PR: per the issue, it requires adjudicating intentional-vs-drift differences in public kwargs, callback signatures, and gauge-reset ordering, which deserve their own review.

## Testing

- Full `Pkg.test()` suite passes (includes Aqua ambiguity checks, explicit-imports checks, JLArrays GPU-semantics tests, and the extensive existing centered/standardized/zerosum property tests, which pin the merged implementations to the old behavior).
- New tests in `test/centered.jl`: `free_energy_h` exact enumeration / mirror consistency / uncentered-equivalence, `∂free_energy_v`/`∂free_energy_h` vs Zygote, and `regularization_penalty` for `CenteredRBM`.

Closes the step-1/step-2 portion of #170.

https://claude.ai/code/session_014sHejbM9ieMFGvhAqh82KU


src/offset_rbms.jl
```diff
@@ -0,0 +1,192 @@
...
+
+_scale_v(rbm::StandardizedRBM) = rbm.scale_v
+_scale_h(rbm::StandardizedRBM) = rbm.scale_h
+_scale_v(rbm::CenteredRBM) = Ones{eltype(rbm.w)}(size(rbm.visible))
+_scale_h(rbm::CenteredRBM) = Ones{eltype(rbm.w)}(size(rbm.hidden))
+
+_maybe_div(x::AbstractArray, ::Ones) = x
+_maybe_div(x::AbstractArray, s::AbstractArray) = x ./ s
+_maybe_mul(x::AbstractArray, ::Ones) = x
+_maybe_mul(x::AbstractArray, s::AbstractArray) = x .* s
```
cossio, line 23: Why is this needed ? I think the performance gain (if any) would be marginal, and the compiler often is smart enough to figure such things out.

cossio: Good question — the point isn't the scalar divisions (LLVM can indeed fold `x / 1.0`), it's the **array allocation that broadcasting forces**. `x ./ Ones(...)` must materialize a fresh array and make a full pass over `x`; neither the compiler nor FillArrays elides that (FillArrays has no specialization for `Array ./ Ones`). `_maybe_div(x, ::Ones) = x` returns the array itself with zero work.

This matters because the shared methods sit on hot paths where the deleted hand-written `CenteredRBM` code had no divisions at all — the specializations are what keep the refactor performance-neutral for centered models:

- `zerosum!` runs **every training iteration** for Potts models. Without `_maybe_div`, `mean(rbm.w ./ scale_v; dims = 1)` copies the entire weight tensor each time. Measured on a (21, 100) Potts visible × 256 hidden RBM: **0.21 ms / 0.2 MB with the specialization vs 6.7 ms / 4.5 MB without** (~30×).
- `inputs_h_from_v` / `inputs_v_from_h` run **every Gibbs step**. The trailing `_maybe_div(inputs, _scale_h(rbm))` would otherwise add a ~2 MB allocation + extra pass per call at batch 1024 (~120 µs, on top of the matmul).

For `StandardizedRBM` nothing changes (its scales are plain arrays, so the `./` method is chosen, exactly as before). If you'd still rather drop the four helper lines and accept the extra per-iteration copies for `CenteredRBM`, I'm happy to simplify — but I'd recommend keeping them.

PR #192, opened 2026-08-01
Use LogStatFunctions.jl for logmeanexp, logvarexp, logstdexp

### LINKED ISSUE #186: LogStatFunctions.jl
Add https://github.com/cossio/LogStatFunctions.jl/ as a dependency once it's registered.

https://github.com/JuliaRegistries/General/pull/162645

cossio:
Closes #186.

[LogStatFunctions.jl](https://github.com/cossio/LogStatFunctions.jl) is now registered in the General registry ([JuliaRegistries/General#162645](https://github.com/JuliaRegistries/General/pull/162645), v2.2.0), so the local `logmeanexp`/`logvarexp`/`logstdexp` implementations can be replaced by it, as requested in the issue.

## Changes

- Delete the local `logmeanexp`, `logvarexp`, and `logstdexp` definitions from `src/ais.jl`. Per review, they are not re-exported: the names are no longer accessible from the `RestrictedBoltzmannMachines` namespace (they were never public API). LogStatFunctions contains the same implementations.
- Since nothing in `src/` uses these functions, LogStatFunctions is not a dependency of the package itself (Aqua's stale-deps check would flag it). It is added as a direct dependency of the **test** and **docs** projects, the actual consumers.
- Drop the now-unused `logsubexp` import from LogExpFunctions.
- Remove the `logmeanexp`/`logvarexp`/`logstdexp` testsets from `test/ais.jl` (that functionality is tested upstream in LogStatFunctions); the remaining AIS testsets import `logmeanexp` from LogStatFunctions. Also drop a few pre-existing unused imports in that file (`randn!`, `logsumexp`, `@inferred`, `free_energy`).
- The AIS literate example imports `logmeanexp`/`logstdexp` from LogStatFunctions, and the `[`logmeanexp`](@ref)` links in the `ais`/`raise` docstrings now link to the LogStatFunctions.jl repository (the `@ref` targets no longer exist in this module).
- Add a `CHANGELOG.md` entry under Unreleased documenting the removal.

## Testing

- `julia --project=test test/ais.jl`, `test/explicit_imports.jl`, and `test/aqua.jl` all pass locally.
- Full `Pkg.test()` suite run on the final state in progress; will follow up if anything fails.

https://claude.ai/code/session_01NJvsUqABYjeACubUNWJiPc


test/explicit_imports.jl
```diff
@@ -17,6 +17,11 @@ end
     public_accesses_without_legacy_metadata = VERSION < v"1.11" ? (:tail,) : ()
     ExplicitImports.test_explicit_imports(
         RBMs;
+        # logmeanexp, logstdexp, and logvarexp are imported from
+        # LogStatFunctions only so they stay accessible under the
+        # RestrictedBoltzmannMachines namespace; no src code calls them.
```
cossio: They don't need to stay accessible from RBM namespace

cossio: Done in 9d6dfe5: removed the re-export and this ignore entry, so the names are no longer accessible from the RBMs namespace. Since nothing in `src/` uses them anymore, LogStatFunctions is also gone from the package `[deps]` (Aqua's stale-deps check would flag it) and is now a direct dependency of the test and docs projects instead, which are the actual consumers. Also removed the `logmeanexp`/`logvarexp`/`logstdexp` testsets from `test/ais.jl` as requested, and updated the CHANGELOG entry accordingly.

---

CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file. The format
 
 ## Unreleased
 
+- Removed `logmeanexp`, `logvarexp`, and `logstdexp` (never public API). They
```
cossio: don't say "never" .... we only defined public API in v6 when we added `public` keyword

cossio: Reworded in 5e4b01d to "(not marked `public`)".

---

PR #198, opened 2026-08-03
Replace vstack/vwiden with Base.stack

cossio:
Split out of #197 (utility consolidation), part 1 of 5.

Deletes the `vstack`/`vwiden` utilities. The gradient and moment helpers (`∂cgfs`, `∂energy_from_moments`, `moments_from_samples`) now use `Base.stack(...; dims = 1)` directly, with `Vector` arguments so the ChainRules `rrule` for `stack(::AbstractArray)` applies.

The `@declare_layer` keyword constructor instead inlines the equivalent `vcat`-of-reshapes, with a comment explaining why `Base.stack` cannot be used there: ChainRules has no `rrule` for tuple arguments (Zygote must differentiate through layer conversions such as `dReLU(::pReLU)` that call this constructor), and `stack` over a `Vector` loses type inference when the parameters mix container types (a view of `par` alongside a broadcast result), which is common in `shift_fields`, `anneal_zero`, and `_drelu_mixture_moments`. `vcat` satisfies all constraints.

Behavior-preserving. Tested with `test/layers.jl`, `test/pottsgumbel.jl`, `test/util.jl`, and `test/jlarrays.jl` (JLArrays with `allowscalar(false)`, confirming `Base.stack` is GPU-safe); the combined change also passed the full suite locally and on CI in #197.

---

### DIFF
--- a/src/layers/abstractlayer.jl
+++ b/src/layers/abstractlayer.jl
@@ -46,6 +46,13 @@ macro declare_layer(Layer, params)
 
     defaults = [Expr(:kw, name, :($init(T, sz))) for (name, init) in zip(names, inits)]
 
+    # Stack the parameters as rows of `par` via vcat of reshapes rather than
+    # `stack`: unlike `stack`, vcat stays inferable when the arguments mix
+    # container types (e.g. a view of `par` alongside a broadcast result) and
+    # has ChainRules rules, so Zygote can differentiate through layer
+    # conversions such as dReLU(::pReLU) that call this constructor.
+    rows = [:(reshape($name, 1, size($name)...)) for name in names]
+
     return esc(
         quote
             Base.@__doc__ struct $Layer{N, A} <: AbstractLayer{N}
@@ -60,7 +67,7 @@ macro declare_layer(Layer, params)
             end
 
             $Layer(par::AbstractArray) = $Layer{ndims(par) - 1, typeof(par)}(par)
-            $Layer(; $(names...)) = $Layer(vstack(($(names...),)))
+            $Layer(; $(names...)) = $Layer(vcat($(rows...)))
             $Layer(::Type{T}, sz::Dims) where {T} = $Layer(; $(defaults...))
             $Layer(sz::Dims) = $Layer(Float64, sz)
 
--- a/src/layers/gaussian.jl
+++ b/src/layers/gaussian.jl
@@ -46,5 +46,8 @@ end
 function moments_from_samples(layer::Gaussian, data::AbstractArray; wts = nothing)
     x1 = batchmean(layer, data; wts)
     x2 = batchmean(layer, data .^ 2; wts)
-    return stack([x1, x2]; dims = 1)
+    # vcat of reshapes rather than `stack`: for unbatched `data`, x1 can be a
+    # view (batchmean returns `data` itself) while x2 is an Array, and `stack`
+    # over mixed container types does not infer
+    return vcat(reshape(x1, 1, size(x1)...), reshape(x2, 1, size(x2)...))
 end
--- a/src/layers/gaussian.jl
+++ b/src/layers/gaussian.jl
@@ -31,7 +31,7 @@ end
 function ∂cgfs(layer::Gaussian, inputs = 0)
     ∂θ = @. (layer.θ .+ inputs) / abs(layer.γ)
     ∂γ = @. -inv(layer.γ) / 2 - sign(layer.γ) * ∂θ^2 / 2
-    return vstack((∂θ, ∂γ))
+    return stack([∂θ, ∂γ]; dims = 1)
 end
 
 function ∂energy_from_moments(layer::Gaussian, moments::AbstractArray)
@@ -40,11 +40,11 @@ function ∂energy_from_moments(layer::Gaussian, moments::AbstractArray)
     x2 = @view moments[2, ..]
     ∂θ = -x1
     ∂γ = @. sign(layer.γ) * x2 / 2
-    return vstack((∂θ, ∂γ))
+    return stack([∂θ, ∂γ]; dims = 1)
 end
 
 function moments_from_samples(layer::Gaussian, data::AbstractArray; wts = nothing)
     x1 = batchmean(layer, data; wts)
     x2 = batchmean(layer, data .^ 2; wts)
-    return vstack((x1, x2))
+    return stack([x1, x2]; dims = 1)
 end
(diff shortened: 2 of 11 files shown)


src/layers/gaussian.jl
```diff
@@ -40,11 +40,11 @@ function ∂energy_from_moments(layer::Gaussian, moments::AbstractArray)
...
     ∂γ = @. sign(layer.γ) * x2 / 2
-    return vstack((∂θ, ∂γ))
+    return stack([∂θ, ∂γ]; dims = 1)
 end
 
 function moments_from_samples(layer::Gaussian, data::AbstractArray; wts = nothing)
     x1 = batchmean(layer, data; wts)
     x2 = batchmean(layer, data .^ 2; wts)
-    return vstack((x1, x2))
+    return stack([x1, x2]; dims = 1)
```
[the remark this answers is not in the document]
cossio: Valid finding, fixed in 79a1198 — verified that `@inferred moments_from_samples(layer, @view samples[:, :, 1])` failed before and passes now. This site now uses the same `vcat`-of-reshapes form as the layer keyword constructor (with a comment explaining why), since it is the other place where mixed container types can reach the stacking: for unbatched `data`, `batchmean` returns the input itself, so `x1` can be a view while `x2` is an `Array`. The other `stack` call sites operate on homogeneous broadcast results, and the `_FieldLayers` single-moment site stacks a one-element vector of a concrete type, so they infer fine.

---

src/layers/gaussian.jl
```diff
@@ -40,11 +40,14 @@ function ∂energy_from_moments(layer::Gaussian, moments::AbstractArray)
...
+    return stack([∂θ, ∂γ]; dims = 1)
 end
 
 function moments_from_samples(layer::Gaussian, data::AbstractArray; wts = nothing)
     x1 = batchmean(layer, data; wts)
     x2 = batchmean(layer, data .^ 2; wts)
-    return vstack((x1, x2))
+    # vcat of reshapes rather than `stack`: for unbatched `data`, x1 can be a
+    # view (batchmean returns `data` itself) while x2 is an Array, and `stack`
+    # over mixed container types does not infer
```
cossio, line 51: you are sure about this ? You tested it?

src/layers/abstractlayer.jl
```diff
@@ -46,6 +46,13 @@ macro declare_layer(Layer, params)
 
     defaults = [Expr(:kw, name, :($init(T, sz))) for (name, init) in zip(names, inits)]
 
+    # Stack the parameters as rows of `par` via vcat of reshapes rather than
+    # `stack`: unlike `stack`, vcat stays inferable when the arguments mix
+    # container types (e.g. a view of `par` alongside a broadcast result) and
+    # has ChainRules rules, so Zygote can differentiate through layer
+    # conversions such as dReLU(::pReLU) that call this constructor.
```
cossio, line 53: you sure about this ? you tested it?

src/layers/gaussian.jl
cossio: Yes, tested — and now committed as a regression test (6b79b8d, `test/layers.jl` "moments_from_samples infers for view data"), so CI enforces it. Concretely, on this branch:

```julia
layer = Gaussian((3, 2)); samples = randn(3, 2, 5); v = @view samples[:, :, 1]
batchmean(layer, v) === v                    # true → x1 is a SubArray
typeof(batchmean(layer, v .^ 2))             # Matrix{Float64} → x2 is an Array

stackform(a, b) = stack([a, b]; dims = 1)
vcatform(a, b)  = vcat(reshape(a, 1, size(a)...), reshape(b, 1, size(b)...))
Base.promote_op(stackform, typeof(v), Matrix{Float64})  # Any
Base.promote_op(vcatform,  typeof(v), Matrix{Float64})  # Array{Float64, 3}
```

With the `stack` form, `@inferred moments_from_samples(layer, v)` threw `return type Array{Float64, 3} does not match inferred return type Any`; with the `vcat` form it passes.

---

src/layers/abstractlayer.jl
cossio: Yes — both claims came out of test failures I hit while iterating on this PR, and both are enforced by the existing suite:

**No ChainRules rrule for tuples.** ChainRules defines `rrule(::typeof(stack), X::AbstractArray; dims)` only (`rulesets/Base/array.jl`), so with a tuple Zygote differentiates through `Base._dim_stack!`, which mutates:

```julia
julia> Zygote.gradient(a -> sum(stack((a, 2a); dims = 1)), ones(3))
ERROR: Mutating arrays is not supported -- called copyto!(SubArray{...}, ...)

julia> Zygote.gradient(a -> sum(stack([a, 2a]; dims = 1)), ones(3))  # Vector → rrule applies
([3.0, 3.0, 3.0],)
```

With a tuple-`stack` constructor, `test/layers.jl:107` (`Zygote.gradient(j -> sum(cgfs(layer, j)), x)`) fails for pReLU/xReLU/nsReLU with exactly that error, because their statistics go through `dReLU(::pReLU)` etc., which call this constructor.

**Vector form loses inference on mixed containers.** `_drelu_mixture_moments` builds `ReLU(; θ = -layer.θn, γ = layer.γn)` — a `Matrix` alongside a `SubArray` — and `Base.vect` promotes that pair to `Vector{AbstractMatrix}`:

```julia
stackform(a, b) = stack([a, b]; dims = 1)
Base.promote_op(stackform, Matrix{Float64}, typeof(l.γn))  # Any
```

With a vector-`stack` constructor, the `@inferred mean_from_inputs(layer)` / `var_from_inputs` tests at `test/layers.jl:55-56` fail for dReLU with `inferred return type Any`. `vcat` of reshapes passes both: it has ChainRules rules and infers `Array{T, N}` for mixed argument containers. Same-typed `shift_fields` / `anneal_zero` constructor calls are also covered by the suite (gauge and AIS tests).

---

test/layers.jl
```diff
@@ -572,3 +572,17 @@ end
     @test mean_abs_from_inputs(drelu) ≈ mean_abs_from_inputs(relu) rtol = 1.0e-4
     @test cgfs(drelu) ≈ cgfs(relu) .+ log1p.(exp.(relu_cgf.(0, 1.0e10) .- cgfs(relu))) rtol = 1.0e-6
 end
+
+using RestrictedBoltzmannMachines: moments_from_samples
```
cossio: always put `using`s in test files at the top of the file

cossio: Moved to the top import block in 1670b3d.

---

PR #200, opened 2026-08-03
Use Base.hypot in sqrt1half

cossio:
Split out of #197 (utility consolidation), part 3 of 5.

`sqrt1half(x)` computes `sqrt(1 + (x/2)^2) + |x|/2`. `Base.hypot` handles the large-argument overflow that previously required an eps-based guard branch, so `_sqrt1half` is deleted and the function body becomes `hypot(1, |x|/2) + |x|/2`.

Behavior-preserving: the existing edge-case tests (`0`, `±1`, `NaN`, `±Inf`, `1e300`, `Float32` type preservation) in `test/truncnorm.jl` all pass unchanged. The combined change also passed the full suite locally and on CI in #197.

---

### DIFF
--- a/src/util/truncated_normal.jl
+++ b/src/util/truncated_normal.jl
@@ -43,8 +43,8 @@ end
 Accurate computation of sqrt(1 + (x/2)^2) + |x|/2.
 """
 function sqrt1half(x::Real)
-    h = abs(float(x)) / 2
-    return hypot(one(h), h) + h
+    h = x / 2
+    return hypot(one(h), h) + abs(h)
 end
 
 """
--- a/src/util/truncated_normal.jl
+++ b/src/util/truncated_normal.jl
@@ -42,14 +42,9 @@ end
 
 Accurate computation of sqrt(1 + (x/2)^2) + |x|/2.
 """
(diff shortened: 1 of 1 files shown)


src/util/truncated_normal.jl
```diff
@@ -42,14 +42,9 @@ end
...
-
-function _sqrt1half(x::Real)
-    if x > 2 / sqrt(eps(x))
-        return x
-    else
-        return sqrt(one(x) + (x / 2)^2) + x / 2
-    end
+function sqrt1half(x::Real)
+    h = abs(float(x)) / 2
+    return hypot(one(h), h) + h
```
cossio: Can we simplify this to:

```
h = x/2
hypot(one(h), h) + abs(h)
```

cossio: Done in 4317687, exactly as suggested. It's equivalent: `x / 2` already produces a float for integer input, `hypot` takes the absolute value internally so the signed `h` is fine, and `one(h)` keeps the width (Float32 stays Float32). Edge cases (`0`, `±1`, `NaN`, `±Inf`, `1e300`, Float32 type preservation) re-verified.

---

PR #201, opened 2026-08-03
Replace the inf/two type-preserving helpers with Base.oftype

cossio:
Split out of #197 (utility consolidation), part 4 of 5.

Deletes the `inf`/`two` convenience helpers in favor of `Base.oftype` at their only two call sites:

- `relu_energy`: `inf(E)` → `oftype(E, Inf)`.
- `tnmean`: `two(a)` → a local `t = oftype(float(a), 2)`.

`oftype(x, v)` is `convert(typeof(x), v)`, which is exactly what the helpers did with a value argument, so value semantics are identical (verified for `Int`, `Float32`, `Float64`, and `BigFloat` arguments). Tests of the deleted helpers are removed.

Tested with `test/util.jl`, `test/truncnorm.jl`, and `test/layers.jl`; the combined change also passed the full suite locally and on CI in #197.

---

### DIFF
--- a/src/util/truncated_normal.jl
+++ b/src/util/truncated_normal.jl
@@ -72,7 +72,7 @@ Mean of the standard normal distribution,
 truncated to the interval (a, +∞).
 """
 function tnmean(a::Real)
-    two = oftype(float(a), 2)
+    two = oftype(a, 2)
     return sqrt(two / π) / erfcx(a / sqrt(two))
 end
 
--- a/src/util/truncated_normal.jl
+++ b/src/util/truncated_normal.jl
@@ -72,8 +72,8 @@ Mean of the standard normal distribution,
 truncated to the interval (a, +∞).
 """
 function tnmean(a::Real)
-    t = oftype(float(a), 2)
-    return sqrt(t / π) / erfcx(a / sqrt(t))
+    two = oftype(float(a), 2)
+    return sqrt(two / π) / erfcx(a / sqrt(two))
 end
 
 """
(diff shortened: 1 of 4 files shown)


src/util/truncated_normal.jl
```diff
@@ -71,7 +71,10 @@ randnt_half(μ::Real, σ::Real) = randnt_half(default_rng(), μ, σ)
 Mean of the standard normal distribution,
 truncated to the interval (a, +∞).
 """
-tnmean(a::Real) = sqrt(two(a) / π) / erfcx(a / √two(a))
+function tnmean(a::Real)
+    t = oftype(float(a), 2)
```
cossio: better to call it `two` instead of `t`

cossio: Also, why `float(a)` instead of just `oftype(a, 2)` ?

cossio: Both done: renamed to `two` in 4bc60a6, and switched to `oftype(a, 2)` in 0565adc. You're right that the `float` was unnecessary — `oftype(a, 2)` is exactly what the old `two(a)` helper did (`convert(typeof(a), 2)`), so it's the more faithful replacement; the only divergence of the `float` variant was for exotic inputs like `Bool`, where the old code threw `InexactError` and `float` would have silently worked. Verified the new form gives identical results to the old helper across `Float64`/`Float32`/`Int`/`Int8`/`BigFloat` arguments.

---

PR #202, opened 2026-08-03
Use Base.stack in collect_states

cossio:
Split out of #197 (utility consolidation), part 5 of 5.

`collect_states` accumulated the enumerated layer states with a `reduce`-with-`cat`, which repeatedly reallocates and is quadratic in the number of states. A single `Base.stack` + `reshape` is linear and keeps the same output shape, eltype, and state order (verified against the old implementation for `Binary` and `Spin` layers).

Tested with `test/rbm.jl` and a direct equivalence check against the previous implementation; the combined change also passed the full suite locally and on CI in #197.

---

### DIFF
--- a/src/partition.jl
+++ b/src/partition.jl
@@ -53,7 +53,5 @@ Only defined for discrete layers.
     For large layers, the exponential number of states will not fit in memory.
 """
 function collect_states(layer::Union{Binary, Spin})
-    return reduce(iterate_states(layer)) do x, y
-        cat(x, y; dims = ndims(layer) + 1)
-    end
+    return reshape(stack(iterate_states(layer)), size(layer)..., :)
 end
--- a/src/partition.jl
+++ b/src/partition.jl
@@ -53,5 +53,8 @@ Only defined for discrete layers.
     For large layers, the exponential number of states will not fit in memory.
 """
 function collect_states(layer::Union{Binary, Spin})
-    return reshape(stack(iterate_states(layer)), size(layer)..., :)
+    states = iterate_states(layer)
+    # explicit state count rather than `:`, which a zero-sized layer would
+    # resolve to 0 and drop its single (empty) state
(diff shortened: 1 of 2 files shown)


src/partition.jl
```diff
@@ -53,7 +53,5 @@ Only defined for discrete layers.
     For large layers, the exponential number of states will not fit in memory.
 """
 function collect_states(layer::Union{Binary, Spin})
-    return reduce(iterate_states(layer)) do x, y
-        cat(x, y; dims = ndims(layer) + 1)
-    end
+    return reshape(stack(iterate_states(layer)), size(layer)..., :)
```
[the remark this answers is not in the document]
cossio: Valid finding, fixed in eb2144f. One correction to the failure mode: on Julia 1.12 the `reshape` does not throw — the colon resolves to 0, silently returning a `(0, 0)` array and dropping the single empty state (which would have made `log_partition` return `-Inf` for a zero-sized visible layer). The fix passes the explicit state count, `reshape(stack(states), size(layer)..., length(states))`, restoring the previous `(0, 1)` result, and adds a regression test for `Binary((0,))` and `Spin((0,))`.

---

src/partition.jl
```diff
@@ -53,7 +53,8 @@ Only defined for discrete layers.
     For large layers, the exponential number of states will not fit in memory.
 """
 function collect_states(layer::Union{Binary, Spin})
-    return reduce(iterate_states(layer)) do x, y
-        cat(x, y; dims = ndims(layer) + 1)
-    end
+    states = iterate_states(layer)
+    # explicit state count rather than `:`, which a zero-sized layer would
+    # resolve to 0 and drop its single (empty) state
```
cossio: this comment is not needed here

cossio: Removed in 24926e6.

---

PR #205, opened 2026-08-07
Replace grad2ave/grad2var with a moments-based layer statistics API

cossio:
## Summary

Replaces the `grad2ave` / `grad2var` gradient-extraction helpers with a moments-based formulation of the layer statistics and of `∂free_energy`, removing them entirely.

### The problem

`grad2ave(layer, ∂)` read the conditional mean out of a cgf-gradient array by exploiting the exponential-family identity `⟨x⟩ = ∂Γ/∂θ`, with a layer-specific slice rule. It existed to share computation inside `∂free_energy` (which needs both `∂Γ` and `⟨h|v⟩`), and its sibling `grad2var` had to invert the reparameterization chain rules, culminating in a 10-line formula for `xReLU`. The name (`2` = "to") also ran against the package-wide `X_from_Y` convention.

### The refactor

`∂cgfs` and `∂energy_from_moments` were near-duplicates: each per-layer `∂cgfs` computed the conditional moments and then applied the (negated) same chain rule that `∂energy_from_moments` applies to sample moments. This PR unfuses those two steps:

- **`moments_from_inputs(layer, inputs)`** (new): per-configuration conditional moments, in the same canonical layout as `moments_from_samples` (first axis = moment index, then `size(layer)`, then batch dimensions). For `pReLU`/`xReLU`/`nsReLU` it is a one-line delegation to `dReLU`, like the rest of their statistics.
- **`∂cgfs`** is now a single generic method, `-∂energy_from_moments ∘ moments_from_inputs` (the cgf/energy conjugacy). All the hand-written per-layer `∂cgfs` chain rules are deleted. `∂energy_from_moments` accepts trailing batch dimensions to support this.
- **`∂free_energy`** becomes symmetric: both layer gradients go through `∂energy_from_moments` — evaluated at the data moments for the visible layer and at the (batch-averaged) conditional moments for the hidden layer — and the per-sample `⟨h|v⟩` needed for `∂w` is read off the conditional moments with `mean_from_moments`. Computed gradients are unchanged (validated against Zygote AD by the existing tests).
- **`mean_from_moments` / `var_from_moments`** (new, replacing `grad2ave` / `grad2var`): trivial reads from the canonical moments layout. The inverse-chain-rule formulas of `grad2var` for `pReLU`/`xReLU` collapse to `xp2 + xn2 - (xp1 + xn1)^2`, and — unlike `grad2var` — the variance is now also recoverable for `nsReLU`, since the dReLU-layout moments carry the second moments.
- **Centered PCD**: the last `grad2ave` call site read `⟨h⟩_d` off the minibatch free-energy gradient — a shortcut the standardized trainer never had (it must recompute a forward pass anyway to get the variance for `scale_h`). The centered trainer now updates its offsets through `center_hidden_from_data!` (which gained a `damping` keyword), symmetric with `standardize_hidden_from_v!`. This costs one extra input matmul per minibatch in centered `pcd!`, the same cost the standardized trainer already pays, and removes the last dependency on gradient-layout introspection.

`src/from_grad.jl` is deleted; the new statistics live in `src/layers/moments.jl`.

Rebased on master after #206 (the GPU pseudolikelihood fix that CI here first surfaced) was merged; the rebased tree is identical to the state that previously passed the full CI matrix.

## Test plan

- [x] `test/layers.jl`, `test/pottsgumbel.jl`, `test/layers/nsReLU.jl`, `test/prelu_validation.jl` updated to the moments API (plus a new per-layer property test comparing `mean_from_moments`/`var_from_moments` of `moments_from_inputs` against `mean_from_inputs`/`var_from_inputs` across batch shapes)
- [x] `test/rbm.jl` passes — `∂free_energy` / `∂free_energy_h` validated against Zygote AD (`@inferred`, weighted and unweighted)
- [x] `test/layers.jl`, `test/pottsgumbel.jl`, `test/layers/nsReLU.jl`, `test/prelu_validation.jl`, `test/centered.jl`, `test/standardized.jl`, `test/zero_weight_training.jl`, `test/jlarrays.jl`, `test/explicit_imports.jl` pass
- [x] Full `Pkg.test()` suite passes locally on this exact tree
- [x] Runic formatting check passes on all changed files

https://claude.ai/code/session_01NK3Q9C91JFmBZxozGJw4wk


cossio:
The CI failure on 14fd80f (`test/jlarrays.jl`, pseudolikelihood testset: "Illegal conversion of a JLArray to a Ptr") is **pre-existing on master**, not caused by this PR's diff: it reproduces identically at the base commit 0e090d6 with the same dependency versions (JLArrays 0.3.2, GPUArrays 11.5.9, Julia 1.12.6).

Root cause: `log_pseudolikelihood_exact` for `Potts` visible layers fed strided views (`view(wflat, rows, :)` etc.) to matrix products. On Julia 1.12 LinearAlgebra dispatches those to BLAS `gemm!` (via the new `_syrk_herk_gemm_wrapper!` path), which needs raw pointers — illegal for GPU arrays, and bypassing the GPUArrays generic matmul.

Fixed in 7776231 by materializing the per-site blocks (indexed copies instead of views) in both the generic Potts path and the Potts/Gaussian specialization — the copies are `q`-row slices, negligible next to the matmuls they feed. Verified locally: the failing testset passes after the fix, and the (previously untested) Potts+Gaussian exact path also matches CPU on JLArrays. Happy to split this into its own PR if you'd rather keep it separate.

---

PR #206, opened 2026-08-07
Fix GPU exact pseudolikelihood for Potts visible layers

cossio:
## Summary

`log_pseudolikelihood(rbm, v; exact = true)` with a `Potts` visible layer errors on GPU arrays with **"Illegal conversion of a JLArray to a Ptr"**. The per-site loops of `log_pseudolikelihood_exact` fed `view`s of the flattened weights/data into matrix products; those views are strided, so LinearAlgebra on Julia 1.12 dispatches `*` to BLAS `gemm!`, which requires raw pointers — illegal for GPU arrays. The GPU generic-matmul hooks are bypassed because the SubArray wrapper satisfies the strided dispatch path.

The failure is deterministic and reproduces on current master (0e090d6) via the `pseudolikelihood` testset of `test/jlarrays.jl`.

**Fix:** materialize the per-site blocks with indexed copies (`wflat[rows, :]` instead of `view(wflat, rows, :)`, etc.) before the matrix products, in both the generic `Potts` path and the `Potts`/`Gaussian` specialization. The copies are `q`-row slices, negligible next to the matmuls they feed. CPU results are the same computation.

Split out of #205 (where CI first surfaced it) so it can be reviewed and merged independently; #205 will be rebased on master once this lands.

## Test plan

- [x] `test/jlarrays.jl` passes on this branch (its `pseudolikelihood` testset fails on master), with `allowscalar(false)`
- [x] Manual JLArrays check of the `Potts`+`Gaussian` exact-pseudolikelihood path (not covered by the testset): matches CPU
- [x] `test/pseudolikelihood.jl` passes (CPU coverage of both modified functions)
- [x] Runic formatting check passes

https://claude.ai/code/session_01NK3Q9C91JFmBZxozGJw4wk

---
_Generated by [Claude Code](https://claude.ai/code/session_01NK3Q9C91JFmBZxozGJw4wk)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file. The format 
 
 ## Unreleased
 
+- Fixed `log_pseudolikelihood(rbm, v; exact = true)` with `Potts` visible
+  layers erroring on GPU arrays ("illegal conversion to a `Ptr`"): the per-site
+  blocks fed to matrix products are now indexed copies instead of views, since
+  strided views of GPU arrays dispatch `*` to BLAS.
+
 ## 6.1.0
 
 - Removed `logmeanexp`, `logvarexp`, and `logstdexp` (not marked `public`). They
--- a/src/pseudolikelihood.jl
+++ b/src/pseudolikelihood.jl
@@ -231,9 +231,11 @@ function log_pseudolikelihood_exact(rbm::RBM{<:Potts}, v::AbstractArray)
     Inew = similar(Iflat)
     for j in 1:nsites
         rows = ((j - 1) * q + 1):(j * q)
-        Vsite = with_eltype_of(wflat, view(vflat, rows, :)) # q × B, one-hot columns
-        θsite = view(θflat, rows) # q
-        Wsite = view(wflat, rows, :) # q × M
+        # indexed copies, not views: strided views of GPU arrays send `*` to
+        # BLAS, which requires raw pointers
+        Vsite = with_eltype_of(wflat, vflat[rows, :]) # q × B, one-hot columns
+        θsite = θflat[rows] # q
+        Wsite = wflat[rows, :] # q × M
         Wold = Wsite' * Vsite # M × B
         θold = Vsite' * θsite # B
         for x in 1:q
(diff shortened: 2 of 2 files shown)


cossio:
Why CI and tests weren't failing before if this bug was already there ?

cossio:
Because the test environment moved underneath the unchanged code: the trigger is the **JLArrays v0.3.2 release**, which landed between master's last CI run and the first CI run of #205. CI resolves the latest compatible versions on every run, so the same commit that was green on Aug 3 fails today.

Evidence:

- Master's last CI run ([Aug 3, on 0e090d6, green](https://github.com/cossio/RestrictedBoltzmannMachines.jl/actions/runs/30848190568)) used Julia 1.12.6 + **JLArrays v0.3.1** + GPUArrays v11.5.8 (from its job log).
- The failing run of #205 (Aug 7) and my local reproduction use the same Julia 1.12.6 but **JLArrays v0.3.2** (+ GPUArrays v11.5.9).
- Bisecting on unfixed master (0e090d6): JLArrays **0.3.1** + GPUArrays **11.5.9** → `log_pseudolikelihood(...; exact = true)` on JLArrays **passes**; JLArrays **0.3.2** → fails. So GPUArrays is innocent; the JLArrays minor release is what changed.

Mechanism: with 0.3.2, `SubArray`s of a `JLArray` now satisfy LinearAlgebra's strided dispatch, so `Wsite' * Vsite` routes to BLAS `gemm!`, which calls `unsafe_convert(Ptr, ...)` — that's JLArrays' own "Illegal conversion of a JLArray to a Ptr" guard. With 0.3.1 the same product fell through to the GPUArrays generic matmul. In other words, the old code worked by dispatch luck on the wrapper type, and the new JLArrays surfaced the latent assumption. The fix here (indexed copies feeding the matmuls) removes that reliance and is backend-agnostic.

It may also be worth an upstream issue at JLArrays.jl asking whether view-matmuls should fall back to the generic GPU path instead of erroring (real CUBLAS handles strided device views, so JLArrays erroring here arguably over-restricts the emulation) — happy to file one if you want.

---

cossio:
One correction to the mechanism above, after diffing the JLArrays 0.3.1 → 0.3.2 sources: on 0.3.1 the view products did **not** fall back to the generic GPU matmul — they also dispatched to BLAS, but `Base.unsafe_convert(Ptr, ::JLArray)` silently returned a pointer into the host-backed buffer, so they ran **CPU BLAS on the emulation array**. Numerically correct, but exactly the implicit CPU `ccall` JLArrays exists to catch. v0.3.2 deliberately added a guard rejecting implicit pointer conversion ("Reject implicit conversions to a pointer, i.e., by passing to a CPU ccall"), which is what surfaced our latent reliance on it. So the JLArrays change is an intentional strictness improvement doing its job, and the fix in this PR is the right response on our side either way (on real CUDA the old views would have worked via CUBLAS's strided support, so this was an emulator-only gap).

I've drafted the upstream issue for JuliaGPU/GPUArrays.jl (JLArrays lives there under `lib/JLArrays`); this session's GitHub access is scoped to this repo, so I've handed the draft to @cossio to file.

---

PR #212, opened 2026-08-08
Unify training and statistics on a single weighted code path

cossio:
## Summary

Every previously unweighted code path is replaced by the weighted one, with `wts` defaulting to lazy uniform weights (`FillArrays.Ones{Bool}`) — no `isnothing`/`isa` control flow and no `wts = nothing` anywhere. Incorporates #218 (closed in favor of this PR): internal weight rescaling is removed entirely. StatsBase was evaluated as a replacement for `wmean` and rejected: its weighted mean fails on GPU arrays (scalar indexing) and under Zygote (internal mutation), both verified; the stdlib matmul composition used here is the ecosystem-standard GPU/AD-safe route and needs no dependency.

- **`wmean(A; wts) = wsum(A, wts) / sum(wts)`**: weights act on the batch dimensions, which are always the trailing dimensions of `A`, so the reduced dimensions are inferred from `size(wts)` (no `dims` keyword; a full reduction is the all-dims case; reduced dimensions are dropped). The reduction is a matmul (`reshape(A, :, n) * vec(wts)`), a plain `sum(A; dims)` for `Ones` weights; for uniform weights `sum(Ones{Bool}(n)) = n::Int`, so `wmean` degenerates to `mean` exactly, without promoting eltypes.
- **`batchmean` always routes through `wmean`** — no unbatched special case; weights always multiply and divide. A scalar `inputs` has no batch dimensions (`batch_size(layer, ::Number) = ()`), flowing through the same path with 0-dimensional `Ones`.
- **Weighted outer products (`∂w` gradient, `batchcov`) are Diagonal-weighted matmuls** via a shared helper: `_weighted_outer(A, wts, B) = A * Diagonal(vec(wts)) * B'`, with lazy `Ones` weights reducing to the plain product (a `Diagonal` around a lazy CPU fill takes a scalar-indexing kernel on GPU arrays, verified with JLArrays under `allowscalar(false)`).
- Non-float operands are floated before the weighted reductions (`_asfloat`, a no-op for float arrays), so narrow-integer data/weights accumulate in float and reduce bit-identically to their float-converted copies (required by the integer/Bool moments testset from #217).
- `wts` defaults to lazy uniform `Ones` **everywhere** — statistics kernels, `initialize!`, and the three `pcd!` trainers. **Breaking**: `wts = nothing` is no longer accepted.
- **Weights must be finite, positive reals**, validated once per run before any mutation; zero or negative weights raise `ArgumentError`, so observations meant to be excluded must be dropped (with their weights) beforehand. Valid weights are used exactly as given: never rescaled (adopted from #218). Extreme weights (near `floatmax`, or needing wider-than-`Float64` accumulation) can overflow the plain reductions; ordinary weights are unaffected.

## Behavioral changes

1. **Breaking**: `pcd!` and `initialize!` require `wts::AbstractVector` (lazy uniform default); `nothing` is rejected.
2. **Breaking**: weights must be strictly positive — zero weights, previously accepted, now raise `ArgumentError` (as do negative, non-finite, or complex weights), before any computation touches the model. Kernels (`wmean`, `∂free_energy`) remain plain unvalidated weighted reductions.
3. **Callbacks** receive lazy `Ones` slices as `wd` for unweighted training (previously `nothing`); weighted-training callbacks receive the raw minibatch weights, as before.
4. Unweighted `batchsize > nsamples` now clamps the batchsize instead of silently performing zero training iterations.
5. The extreme-weight scale-stability guarantee and its tests are removed.

Also fixed in passing: `batchcov` on unbatched data.

## Tests

Updated for the new contracts (`test/util.jl`, `test/zero_weight_training.jl`, `test/infinite_minibatches.jl`, `test/layers.jl`, `test/prelu_validation.jl`, `test/standardized.jl`, `test/jlarrays.jl`). Full suite green locally on Julia 1.12.6 at each review round, including JLArrays with `allowscalar(false)`.

## Version

Contains breaking API changes (`wts = nothing` rejected, zero weights rejected by public trainers); suggest **7.0.0** per ColPrac — final call is the owner's, `Project.toml` untouched.

https://claude.ai/code/session_01UxgabFEUK6LnGrjHKXupAJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01UxgabFEUK6LnGrjHKXupAJ)_


src/util/util.jl
```diff
@@ -1,40 +1,65 @@
...
+an ordinary `mean` without allocating a weights array or promoting eltypes.
+"""
+function wmean(A::AbstractArray; dims = :, wts::AbstractArray = _uniform_wts(A, dims))
+    return wsum(A, wts; dims) / sum(wts)
 end
 
-_weighted_term(a, w) = iszero(w) ? zero(a) * w : a * w
+# Lazy uniform weights matching the reduced dimensions `dims` of `A`.
+_uniform_wts(A::AbstractArray, ::Colon) = Ones{Bool}(size(A))
+_uniform_wts(A::AbstractArray, dims) = Ones{Bool}(ntuple(i -> size(A, dims[i]), length(dims)))
```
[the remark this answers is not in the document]
cossio: This isn't the case in Julia: numbers are iterable and indexable, so `length(2) == 1` and `(2)[1] == 2`, and `_uniform_wts(A, 2)` returns `Ones{Bool}(size(A, 2))` as intended. `wmean(A; dims = 2)` with the default weights is exercised directly by the "wmean is a plain weighted mean" testset in `test/zero_weight_training.jl` (and `dims = 4` with explicit weights in `test/util.jl`), and both pass. No change needed.

---

src/train/infinite_minibatches.jl
```diff
@@ -54,51 +39,62 @@ function Base.iterate(iter::InfiniteMinibatchIterator, (i, shuffled))
...
+    return getobs(positive_indices, data, wts)
 end
 
-_batch_weight(::Nothing, ::Nothing) = 1
+# Normalizing by the largest weight bounds the prepared weights by one, so
+# extreme finite weights cannot overflow the training-loop reductions. The
+# division stays in `float(eltype(wts))` to avoid promoting the arrays the
+# weights later multiply; uniform `Ones` weights stay lazy.
+_normalize_weights(wts::Ones) = wts
+_normalize_weights(wts::AbstractVector) = wts ./ maximum(wts)
```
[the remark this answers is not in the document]
cossio: Good catch — fixed in 55941b2. `_prepare_training_data` now validates first, then normalizes, then filters, so relative weights that underflow `float(eltype(wts))` are dropped exactly like zero weights and no minibatch can end up with weights summing to zero. Covered by a new `Float16[nextfloat(0), floatmax]` case in the "training weights are prepared once per run" testset.

---

src/util/util.jl
```diff
@@ -1,40 +1,65 @@
...
+
+```math
+\frac{\sum_i A_i w_i}{\sum_i w_i}
+```
+
+`wts` defaults to lazy uniform weights (`FillArrays.Ones`), which reduce like
+an ordinary `mean` without allocating a weights array or promoting eltypes.
+"""
+function wmean(A::AbstractArray; dims = :, wts::AbstractArray = _uniform_wts(A, dims))
+    return wsum(A, wts; dims) / sum(wts)
```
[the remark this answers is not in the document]
cossio: Fixed in 55941b2: `initialize!` (RBM-level and per-layer) and `initialize_w!` now run the same `_validate_weights` + `_normalize_weights` hygiene as training on explicit weights, so extreme finite weights are normalized by their maximum before any reduction and cannot overflow. New "initialize! weight hygiene" testset checks `floatmax(Float64)` weights produce the same finite model as unit weights.

---

src/train/initialization.jl
```diff
@@ -19,17 +19,24 @@ function initialize!(rbm::RBM; ϵ::Real = 1.0e-6)
...
     initialize!(rbm.hidden)
     initialize_w!(rbm, data; ϵ, wts)
     zerosum!(rbm)
     return rbm
 end
 
-function initialize!(layer::Binary, data::AbstractArray; ϵ::Real = 1.0e-6, wts = nothing)
+function initialize!(
+        layer::Binary, data::AbstractArray;
+        ϵ::Real = 1.0e-6, wts::AbstractArray = uniform_weights(layer, data)
```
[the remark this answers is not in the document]
cossio: Fixed in 55941b2: all data-taking layer `initialize!` methods accept `wts = nothing` again (`Union{AbstractArray, Nothing} = nothing`), resolved by the `_initialization_weights` dispatch helper to lazy uniform weights. Tested in the new "initialize! weight hygiene" testset.

---

src/train/train.jl
```diff
@@ -36,21 +36,25 @@ function _train!(
...
+        batch_weight = _batch_weight(wd, wts_mean)
 
         # positive and negative phase gradients
         ∂d = ∂free_energy(rbm, vd; wts = wd, moments)
         ∂m, extras = negative_phase(vd)
-        ∂ = (∂d - ∂m) * batch_weight # correct weighted minibatch bias
+        # Correct the weighted minibatch bias. The correction is a Float64
+        # scalar; convert it to the gradient eltype so it cannot promote
+        # narrow (e.g. Float32) parameters.
+        ∂ = (∂d - ∂m) * convert(float(real(eltype(∂d.w))), batch_weight)
```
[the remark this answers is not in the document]
cossio: Addressed in 55941b2: the conversion now saturates at `floatmax` of the gradient eltype (`convert(T, min(batch_weight, floatmax(T)))`), so an extreme correction cannot round to `Inf` and poison the update. I kept the conversion to the gradient eltype rather than widening: multiplying by a Float64 scalar would promote every Float16/Float32 gradient array to Float64, which is exactly the promotion this PR removes. In the saturating regime (a >65504× correction on Float16 parameters) the update is clamped rather than exact — that skew is already far outside what Float16 training can represent elsewhere (free energies, Adam moments).

---

src/train/infinite_minibatches.jl
```diff
@@ -54,51 +39,62 @@ function Base.iterate(iter::InfiniteMinibatchIterator, (i, shuffled))
...
     length(wts) == size(data, ndims(data)) ||
         throw(DimensionMismatch("length(wts) must equal the number of data samples"))
+    data, wts = _filter_zero_weights(data, wts)
+    wts = _normalize_weights(wts)
+    # Scalar Float64 accumulation; it never touches the array eltypes.
+    wts_mean = sum(Float64, wts) / length(wts)
+    return data, wts, wts_mean, min(batchsize, length(wts))
+end
+
+_filter_zero_weights(data::AbstractArray, wts::Ones) = (data, wts)
```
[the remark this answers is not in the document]
cossio: Fixed in 55941b2: every lazy fast path (`_validate_weights`, `_normalize_weights`, `_filter_zero_weights`, `_scale_obs`, `_wsum_all`, `_wsum_trailing`) is now restricted to `Ones{<:Real}`, so complex-valued `Ones` fall through to the elementwise validation and fail with the promised `ArgumentError` at the boundary. Tested with `Ones{ComplexF64}(2)` in `test/zero_weight_training.jl`.

---

src/util/util.jl
```diff
@@ -1,40 +1,65 @@
...
```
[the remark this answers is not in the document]
cossio: Fixed in e167a3e: `_normalize_weights` now divides in `float(eltype(wts))` widened to at least `Float32`, so `Float16` weights are prepared as `Float32` at the boundary (initialize! and training alike) — their sums can no longer overflow and moderate ratios survive. `Float32` and wider weights keep their eltype, preserving the no-promotion guarantee. Tested with `ones(Float16, 70_000)` and a `Float16[nextfloat(0), floatmax]` ratio-retention case.

---

src/train/train.jl
```diff
@@ -36,21 +36,27 @@ function _train!(
...
         # positive and negative phase gradients
         ∂d = ∂free_energy(rbm, vd; wts = wd, moments)
         ∂m, extras = negative_phase(vd)
-        ∂ = (∂d - ∂m) * batch_weight # correct weighted minibatch bias
+        # Correct the weighted minibatch bias. The correction is a Float64
+        # scalar; convert it to the gradient eltype so it cannot promote
+        # narrow (e.g. Float32) parameters, saturating at the largest finite
+        # value so extreme corrections cannot round to Inf in narrow types.
+        T = float(real(eltype(∂d.w)))
+        ∂ = (∂d - ∂m) * convert(T, min(batch_weight, floatmax(T)))
```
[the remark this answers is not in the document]
cossio: Fixed in e167a3e: the clamp is replaced by `_scale_gradient`, which converts the correction to the gradient eltype when it is representable there (the common case, one scalar conversion, no promotion), and otherwise forms each product in the wide scalar type and converts back (`T.(∂.w .* λ)` — a single fused broadcast, no wide temporary). Representable products stay exact instead of saturating, and no bias is introduced. Unit-tested with Float16 gradients and a 1e5 correction.

---

src/train/initialization.jl
```diff
@@ -19,17 +19,37 @@ function initialize!(rbm::RBM; ϵ::Real = 1.0e-6)
...
 
-function initialize!(rbm::RBM, data::AbstractArray; ϵ::Real = 1.0e-6, wts = nothing)
+#= Public `initialize!` boundaries accept `wts = nothing` (lazy uniform
+weights) and apply the same hygiene as training to explicit weights:
+validation and normalization by the largest weight, so extreme finite weights
+cannot overflow the moment-matching reductions. =#
+_initialization_weights(::Nothing, layer::AbstractLayer, data::AbstractArray) =
+    uniform_weights(layer, data)
+_initialization_weights(wts::AbstractArray, ::AbstractLayer, ::AbstractArray) =
+    _normalize_weights(_validate_weights(wts))
```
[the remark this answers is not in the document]
cossio: Fixed in e167a3e: the `initialize!` boundary (`_initialization_data`) now drops zero-weight samples together with validation and normalization, for the RBM-level method, `initialize_w!`, and every data-taking layer method (vector weights over a trailing batch dimension; multi-dimensional batch weights are left unfiltered). `initialize!(Binary((1,)), [NaN 1.0]; wts = [0.0, 1.0])` now produces finite parameters — tested in the "initialize! weight hygiene" testset.

---

src/train/infinite_minibatches.jl
```diff
@@ -54,51 +39,71 @@ function Base.iterate(iter::InfiniteMinibatchIterator, (i, shuffled))
...
-    return mean(wd ./ normalization.scale) / normalization.mean
+    # GPUArrays do not generally support logical indexing. Transfer only
+    # the mask used to build indices; indexing preserves the data backend.
+    positive_indices = findall(adapt(Array, positive))
+    return getobs(positive_indices, data, wts)
 end
+
+# mean(wd) / mean(wts), the bias correction for a weighted minibatch, as a
+# Float64 scalar (converted to the gradient eltype at its use site).
+_batch_weight(wd::AbstractVector, wts_mean::Float64) = sum(Float64, wd) / (length(wd) * wts_mean)
```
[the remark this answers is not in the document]
cossio: Fixed in e167a3e: `_mean_of_weights` and `_batch_weight` accumulate in `promote_type(Float64, float(eltype(wts)))` — Float64 for ordinary weights, `BigFloat` for `BigFloat` weights — so representable ratios like `big"1e-400"` survive into the correction, which is only converted at the gradient application. Tested: `BigFloat[big"1e-400", 1]` with `batchsize = 1` yields a positive correction ≈ `2e-400`.

---

src/util/util.jl
```diff
@@ -1,40 +1,65 @@
...
         wsz = ntuple(ndims(A)) do i
-            i ∈ dims ? size(A, i) : 1
+            i ∈ rdims ? size(A, i) : 1
         end
-        w = reshape(wts, wsz)
+        return sum(A .* reshape(wts, wsz); dims)
     end
+end
+
+_wsum_all(A::AbstractArray, wts::AbstractArray) = dot(vec(A), vec(wts))
```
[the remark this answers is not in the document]
cossio: Fixed in e167a3e: the full-reduction path uses `transpose(vec(A)) * vec(wts)` instead of `dot`, which does not conjugate and still hits the fast BLAS path for real arrays. `wmean(ComplexF64[1 + im]; wts = [1.0]) == 1 + im` is now tested in `test/util.jl`.

---

src/train/initialization.jl
```diff
@@ -19,17 +19,51 @@ function initialize!(rbm::RBM; ϵ::Real = 1.0e-6)
...
+    wts = _normalize_weights(_validate_weights(wts))
+    return _filter_zero_weights(data, wts)
+end
+
+# Vector weights over a single trailing batch dimension are filtered like
+# training data; other weight shapes (multi-dimensional batches) are left as
+# they are.
+_drop_zero_weight_samples(data::AbstractArray, wts::AbstractVector, layerdims::Int) =
+    ndims(data) == layerdims + 1 ? _filter_zero_weights(data, wts) : (data, wts)
+_drop_zero_weight_samples(data::AbstractArray, wts::AbstractArray, ::Int) = (data, wts)
```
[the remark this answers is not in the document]
cossio: Fixed in 128525d: `_initialization_data` now flattens the trailing batch dimensions (`reshape(data, size(layer)..., :)` with `vec(wts)`) before dropping zero-weight configurations, so filtering applies uniformly to any batch layout — the moment-matching reductions are invariant under that reshape. Your exact example (`reshape([NaN, 1, 1, 1], 1, 2, 2)` with `wts = [0 1; 1 1]`) is now a test case producing finite parameters.

---

src/train/initialization.jl
```diff
@@ -19,17 +19,48 @@ function initialize!(rbm::RBM; ϵ::Real = 1.0e-6)
...
+    # Flatten the trailing batch dimensions, so zero-weight configurations are
+    # dropped uniformly regardless of the batch layout; the moment-matching
+    # reductions are invariant under this reshape.
+    return _filter_zero_weights(reshape(data, size(layer)..., :), vec(wts))
+end
+_initialization_data(::Nothing, data::AbstractArray) =
+    (data, Ones{Bool}(size(data, ndims(data))))
+function _initialization_data(wts::AbstractVector, data::AbstractArray)
+    wts = _normalize_weights(_validate_weights(wts))
+    return _filter_zero_weights(data, wts)
```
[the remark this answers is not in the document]
cossio: Fixed in bcab2b9: `_initialization_data` checks `length(wts) == size(data, ndims(data))` before validating/filtering, so undersized weights raise `DimensionMismatch` (matching `_prepare_training_data`) instead of silently truncating the data. The layer-level variant already asserts `size(wts) == batch_size(layer, data)`. Tested with a 3-sample dataset and 2 weights containing a zero.

---

src/train/infinite_minibatches.jl
```diff
@@ -54,51 +39,82 @@ function Base.iterate(iter::InfiniteMinibatchIterator, (i, shuffled))
...
+    npositive == length(wts) && return data, wts
+
+    # GPUArrays do not generally support logical indexing. Transfer only
+    # the mask used to build indices; indexing preserves the data backend.
+    positive_indices = findall(adapt(Array, positive))
+    return getobs(positive_indices, data, wts)
 end
 
-_batch_weight(::Nothing, ::Nothing) = 1
+# Mean of prepared weights as a scalar in at least Float64 precision (wider if
```
cossio: why we need at least Float64? What was wrong with the previous implementation?

cossio: Two reasons, both cheap because this is a scalar statistic that never touches the array eltypes:

1. **Parity with master.** The old code computed `normalization.mean = mean(wts ./ scale)` where `scale = 1.0 * float(maximum(wts))` — the Float64 scale promoted the broadcast, so this mean was already accumulated in ≥Float64 (wider for BigFloat weights). The new `promote_type(Float64, float(eltype(wts)))` reproduces exactly that width, minus the per-iteration `wd ./ scale` array allocation.
2. **The correction is a ratio of two reductions.** With the weights now prepared in their own (possibly Float32) eltype, accumulating `mean(wd)` and `mean(wts)` in that eltype would let summation error scale with the dataset size into a factor multiplying every gradient update. Float64 makes it exact for free; `promote_type` (rather than plain `Float64`) keeps BigFloat weights from being truncated (a ratio like `big"1e-400"` rounds to zero through Float64, dropping that sample's contribution entirely — one of the Codex findings).

If you'd rather accumulate in `float(eltype(wts))` and accept the rounding, that also works for realistic weights — but there's no cost to the wider scalar, so I kept it.

Separately, following your comment on `_scale_gradient`: removed in 0bb736b. The correction is bounded by the number of samples (the largest prepared weight is 1, so `mean(wts) ≥ 1/n`), so a plain `convert` to the gradient eltype cannot overflow Float32 or wider — the wide-product branch only served Float16 models with >65504 heavily skewed samples, which the package doesn't support; that regime now fails loudly instead.

---

src/train/train.jl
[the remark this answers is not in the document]
cossio: Update: per the repository owner's request, `_scale_gradient` was removed again in 0bb736b in favor of a plain `convert` to the gradient eltype. Rationale: the correction is bounded by the number of samples (after normalization the largest weight is 1, so `mean(wts) ≥ 1/n`), which makes the conversion overflow-free for Float32 and wider gradients; only Float16 models with >65504 heavily skewed samples could overflow, a regime the package does not support — and that now fails loudly (Inf) rather than silently biasing.

---

src/train/infinite_minibatches.jl
cossio: Simplified in 626bfb5: `_mean_of_weights` is gone and both the per-run mean and the per-batch correction are plain `Statistics.mean` in the weights' own precision. Note BigFloat weights still keep their precision naturally (`mean(::Vector{BigFloat})` is a BigFloat), so the earlier Codex BigFloat case still passes; the only change is that narrow-eltype weights now round in their own eltype, which is fine as you say. Tests updated accordingly, zero_weight/train/jlarrays green.

---

src/util/util.jl
```diff
@@ -1,40 +1,66 @@
...
 
-    if dims === (:)
-        @assert size(wts) == size(A)
-        w = wts
+`wts` has the shape of the reduced dimensions of `A` (of all of `A` for
+`dims = :`, returning a scalar). Reduced dimensions are kept as singletons,
+like `sum(A; dims)`.
+"""
+function wsum(A::AbstractArray, wts::AbstractArray; dims = :)
+    dims === (:) && return _wsum_all(A, wts)
```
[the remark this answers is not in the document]
cossio: Fixed in ed9d6fd: the `dims = :` path asserts `size(wts) == size(A)` before dispatching to `_wsum_all`, restoring the previous contract for both ordinary and `Ones` weights. Tested with a shape-mismatched `Ones` and a same-length reshaped weights array.

---

src/train/gradient.jl
```diff
@@ -23,7 +23,7 @@ Gradient of `free_energy(rbm, v)` with respect to model parameters.
 If `v` consists of multiple samples (batches), then an average is taken.
 """
 function ∂free_energy(
-        rbm, v::AbstractArray; wts = nothing,
+        rbm, v::AbstractArray; wts::AbstractArray = uniform_weights(rbm.visible, v),
```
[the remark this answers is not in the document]
cossio: Declining this one as a deliberate design decision of the PR. `∂free_energy`, `batchmean`, `moments_from_samples`, etc. are internal statistics kernels — none are in the package's `public` list — and the whole point of this PR is a single weighted code path with no `Nothing` handling inside the kernels. `nothing` remains accepted at the documented public boundaries (`pcd!`, `initialize!`), where it is translated once by dispatch. Re-adding `Union{AbstractArray, Nothing}` to every kernel would reintroduce the dual-path surface this PR removes; callers forwarding an optional weights variable can go through those boundaries or pass `uniform_weights(layer, x)`. This is called out in the PR description and CHANGELOG for the owner to confirm.

---

src/centered.jl
```diff
@@ -184,19 +184,28 @@ function center_hidden!(rbm::CenteredRBM, offset_h::AbstractArray)
     return rbm
 end
 
-function center_visible_from_data!(rbm::CenteredRBM, data::AbstractArray; wts = nothing)
+function center_visible_from_data!(
+        rbm::CenteredRBM, data::AbstractArray;
+        wts::AbstractArray = uniform_weights(rbm.visible, data)
```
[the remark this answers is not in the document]
cossio: Declining, same rationale as the statistics kernels: `center_*_from_data!` and `standardize_*` are not in the package's `public` list and, in the supported flow, are only invoked by the trainers with weights that already went through `_prepare_training_data` (validated, normalized, zero-filtered) — the per-iteration `post_update!` calls receive prepared minibatch slices, and re-running hygiene there would add per-iteration cost to guard a direct-call corner. The contract — kernels are plain weighted reductions, hygiene lives at the `pcd!`/`initialize!` boundaries — is documented in the PR description for the owner to confirm; the owner has twice asked to remove defensive machinery from these paths, not add more.

---

src/layers/abstractlayer.jl
```diff
@@ -164,14 +164,24 @@ function batch_size(layer::AbstractLayer, x::AbstractArray)
...
-function batchmean(layer::AbstractLayer, x::AbstractArray; wts = nothing)
+function batchmean(
+        layer::AbstractLayer, x::AbstractArray; wts::AbstractArray = uniform_weights(layer, x)
+    )
     @assert size(layer) == size(x)[1:ndims(layer)]
     if ndims(layer) == ndims(x)
-        wts::Nothing
+        @assert length(wts) == 1
         return x
     else
```
cossio: can this logic be handled by `wmean` ? We should in any case multiply and divide by wts . This would for example take care of the scenario where `wts = [NaN]`

cossio: So I think it'd be cleaner if we just called `wmean` in any case.

cossio: Done in 7322f23: `batchmean` always routes through `wmean` — the unbatched special case is gone, so weights always multiply and divide (a `NaN` weight now propagates instead of being ignored). This also let the Gaussian moments `vcat` workaround become a plain `stack`, since unbatched `batchmean` no longer returns the data array itself.

The same commit implements the follow-up simplification you suggested: every `wmean` call reduces trailing dimensions (verified: `batchdims`, `batchmean_moments`, `∂cgf`, and the full-reduction case are all trailing), so `wmean`/`wsum` lost their `dims` keyword — the reduced dimensions are inferred from `size(wts)` matching the trailing dimensions of `A`, the non-trailing broadcast fallback is deleted, reduced dimensions are dropped from the result (removing the reshape at each call site), and `_uniform_wts` is gone.

---

src/util/util.jl
```diff
@@ -1,40 +1,57 @@
...
+        # of `A` and is a plain `sum` for lazy uniform `Ones` weights.
+        S = _wsum_trailing(reshape(A, :, length(wts)), vec(wts))
+        return reshape(S, ntuple(d -> size(A, d), Val(ndims(A) - ndims(wts))))
     end
+end
+
+# `transpose`, not `dot`: the documented sum is `Σ Aᵢwᵢ`, without conjugation
+_wsum_all(A::AbstractArray, wts::AbstractArray) = transpose(vec(A)) * vec(wts)
+_wsum_all(A::AbstractArray, ::Ones{<:Real}) = sum(A)
+_wsum_trailing(A::AbstractMatrix, wts::AbstractVector) = A * wts
```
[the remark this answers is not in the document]
cossio: Declining — the premise doesn't hold for this codebase. Mixed-eltype device matmul goes through GPUArrays' generic kernel (not CUBLAS), which works without scalar indexing: verified with JLArrays under `allowscalar(false)`, both `JLMatrix{Float32} * JLVector{Float64}` and the mat-mat shape. Two further points: (a) master's weighted path already performed exactly this mixed matmul on GPU (`vw::Float64 * hz'::Float32` in `∂interaction_energy`), so this PR adds no new requirement; (b) the weighted JLArrays `pcd!` test (Float32 model, Float64 weights) exercises the full path and passes. Converting operands would either promote a full data array (the allocation this PR removes) or silently change the promoted result eltype.

---

src/train/infinite_minibatches.jl
```diff
@@ -54,51 +39,77 @@ function Base.iterate(iter::InfiniteMinibatchIterator, (i, shuffled))
...
         batchsize::Int,
     )
     batchsize > 0 || throw(ArgumentError("batchsize must be positive"))
-    isnothing(wts) && return data, wts, nothing, batchsize
-
     length(wts) == size(data, ndims(data)) ||
         throw(DimensionMismatch("length(wts) must equal the number of data samples"))
+    wts = _normalize_weights(_validate_weights(wts))
+    data, wts = _filter_zero_weights(data, wts)
+    return data, wts, mean(wts), min(batchsize, length(wts))
```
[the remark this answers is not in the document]
cossio: Declining as pre-existing, deliberate behavior rather than a regression of this PR: on master the weighted path likewise clamped only the iterator after `vm` was built, and the "fewer positive samples than batchsize complete" testset explicitly pins that the default number of fantasy chains is the *requested* batchsize (the chain count is a free parameter of the negative phase, not tied to the minibatch size). This PR only extends the existing clamp to the unweighted case. Whether default chains should track the clamped batchsize instead is a separate design question for the owner — happy to change it in a follow-up if desired.

---

src/util/util.jl
```diff
@@ -1,40 +1,57 @@
...
+[`wsum`](@ref)). By default, lazy uniform weights over all of `A`, which
+reduce like an ordinary `mean` without allocating a weights array or
+promoting eltypes.
+
+```math
+\frac{\sum_i A_i w_i}{\sum_i w_i}
+```
+"""
+function wmean(A::AbstractArray; wts::AbstractArray = Ones{Bool}(size(A)))
+    return wsum(A, wts) / sum(wts)
```
cossio, line 66: I think we should divide by mean(wts) instead of sum(wts). This way, when wts = Ones, wmean reduces to mean. Right ?

src/util/util.jl
```diff
@@ -1,40 +1,57 @@
...
+        # A matmul-shaped reduction, which never materializes a weighted copy
+        # of `A` and is a plain `sum` for lazy uniform `Ones` weights.
+        S = _wsum_trailing(reshape(A, :, length(wts)), vec(wts))
+        return reshape(S, ntuple(d -> size(A, d), Val(ndims(A) - ndims(wts))))
     end
+end
+
+# `transpose`, not `dot`: the documented sum is `Σ Aᵢwᵢ`, without conjugation
+_wsum_all(A::AbstractArray, wts::AbstractArray) = transpose(vec(A)) * vec(wts)
+_wsum_all(A::AbstractArray, ::Ones{<:Real}) = sum(A)
```
cossio: this "optimization" is not really buying us anything, I think

src/util/util.jl
```diff
@@ -1,40 +1,57 @@
...
+function wmean(A::AbstractArray; wts::AbstractArray = Ones{Bool}(size(A)))
+    return wsum(A, wts) / sum(wts)
 end
 
-_weighted_term(a, w) = iszero(w) ? zero(a) * w : a * w
+# Scale the batch columns of `A` by their weights, for matmul-shaped weighted
+# reductions. Fold the weights into the smaller factor of a product to keep the
+# temporary small; lazy uniform `Ones` weights are a no-op.
+_scale_obs(A::AbstractMatrix, wts::AbstractVector) = A .* reshape(wts, 1, :)
+_scale_obs(A::AbstractMatrix, ::Ones{<:Real}) = A
```
cossio: this "optimization" is not really buying us anything, I think

src/train/initialization.jl
```diff
@@ -140,16 +200,15 @@ Initializes `rbm.w` such that typical inputs to hidden units are λ.
 """
 function initialize_w!(
         rbm::RBM, data::AbstractArray;
-        λ::Real = 0.1, ϵ::Real = 1.0e-6, wts::Union{Nothing, AbstractVector} = nothing
+        λ::Real = 0.1, ϵ::Real = 1.0e-6, wts::Union{AbstractVector, Nothing} = nothing
```
cossio: why `wts` can be `Nothing` here ? We want the default to be `Ones`, like everywhere else

src/util/util.jl
```diff
@@ -1,40 +1,57 @@
...
+"""
+function wmean(A::AbstractArray; wts::AbstractArray = Ones{Bool}(size(A)))
+    return wsum(A, wts) / sum(wts)
 end
 
-_weighted_term(a, w) = iszero(w) ? zero(a) * w : a * w
+# Scale the batch columns of `A` by their weights, for matmul-shaped weighted
+# reductions. Fold the weights into the smaller factor of a product to keep the
+# temporary small; lazy uniform `Ones` weights are a no-op.
+_scale_obs(A::AbstractMatrix, wts::AbstractVector) = A .* reshape(wts, 1, :)
```
cossio: I don't think this needs to be a separate method. Let's try to keep things simple

src/centered.jl
```diff
@@ -253,7 +262,7 @@ function pcd!(
         wts::Union{AbstractVector, Nothing} = nothing, # data weights
```
cossio: default wts is Ones, not nothing

src/util/util.jl
cossio: Disregard previous comment

src/util/util.jl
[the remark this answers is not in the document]
cossio: Fixed in 2f2917c: non-float operands are floated before the weighted matmuls (`_asfloat`, a no-op for float arrays), so narrow-integer data and weights accumulate in float — your `ones(Int8, 128)` example now returns `[1.0]` — and integer data reduces bit-identically to its float-converted copy (which also fixes the CI failure in the new integer/Bool moments testset).

---

src/util/util.jl
cossio: With `sum(wts)` the property you want already holds, and more exactly: `sum(Ones{Bool}(n))` is `n::Int`, so for uniform weights `wmean = wsum / n`, which is literally `mean` — bit-for-bit, since the `Ones` path lowers to `sum(A; dims)`. Dividing by `mean(wts)` instead would divide by `mean(Ones{Bool}(n)) = 1.0::Float64`, which promotes Float32 results to Float64 and adds a needless division. So I kept `sum(wts)`: mathematically `Σᵢ wᵢ Aᵢ / Σᵢ wᵢ` is the same quantity either way, and the `sum` form is the one that degenerates to `mean` exactly. Happy to switch if you prefer the `mean`/`mean` reading regardless.

---

src/util/util.jl
cossio: Removed in 2f2917c (the `_wsum_all` `Ones` method). Only visible effect: a full reduction with default weights now returns a Float64 scalar instead of preserving e.g. Float32 — scalars don't promote arrays, so that's harmless.

---

src/util/util.jl
cossio: Removed in 2f2917c (the `_scale_obs` `Ones` no-op; the generic broadcast now runs for uniform weights too, costing one `hflat`-sized temporary per iteration — multiplying by `true` keeps values and eltype, so results are unchanged). I kept the one `Ones` method on `_wsum_trailing` (`sum(A; dims = 2)`): that one carries the unweighted `batchmean`/moments path, where the generic route would add a data-sized copy per call and promote Float32 to Float64 through `float.(Ones{Bool})`. Can drop it too if you'd rather.

---

src/train/initialization.jl
cossio: Done in 2f2917c, and swept everywhere: `initialize_w!`, the whole `initialize!` family, and the three `pcd!` trainers now default `wts` to lazy uniform `Ones` and accept `AbstractVector`/`AbstractArray` only — `wts = nothing` is gone from the package (flagged as breaking in the CHANGELOG since the public trainers previously accepted it).

---

src/util/util.jl
cossio: Done in 2f2917c: `_scale_obs` is gone; the weight broadcast is written inline at its three use sites (`∂interaction_energy`, `batchcov`, `initialize_w!`).

---

src/centered.jl
cossio: Done in 2f2917c for all three `pcd!` trainers: `wts::AbstractVector = Ones{Bool}(size(data, ndims(data)))`, no `Nothing` anywhere.

---

src/layers/abstractlayer.jl
```diff
@@ -164,92 +164,102 @@ function batch_size(layer::AbstractLayer, x::AbstractArray)
 end
 
 """
-    batchmean(layer, x; wts = nothing)
+    uniform_weights(layer, x)
+
+Lazy uniform weights (`FillArrays.Ones`) over the batch dimensions of `x`.
+"""
+uniform_weights(layer::AbstractLayer, x::AbstractArray) = Ones{Bool}(batch_size(layer, x))
+uniform_weights(::AbstractLayer, ::Real) = Ones{Bool}()
```
cossio: why is this method needed?

cossio: It serves the scalar-`inputs` case: `∂cgf(layer, inputs = 0)` and `total_mean_from_inputs`/`total_var_from_inputs`/`total_meanvar_from_inputs(layer, inputs = 0)` default their weights to `uniform_weights(layer, inputs)`, and `inputs` defaults to the scalar `0` (broadcast over the layer, no batch dimensions). For that case this method returns a 0-dimensional `Ones` — "nothing to reduce" — which flows through `batchmean`/`wmean` uniformly (their shape asserts see the empty batch tuple).

Without it, those four functions would need a type test in their default expressions (`inputs isa Real ? ... : ...`). An alternative with the same method count but arguably better placement: define `batch_size(layer, ::Real) = ()` and keep a single generic `uniform_weights(layer, x) = Ones{Bool}(batch_size(layer, x))` covering both — happy to switch if you prefer that reading.

---

src/layers/abstractlayer.jl
```diff
@@ -164,92 +164,102 @@ function batch_size(layer::AbstractLayer, x::AbstractArray)
...
-        C = ξ * ξ' / size(ξ, 2)
-    else
-        @assert size(wts) == batch_size(layer, x)
-        w = Diagonal(vec(wts))
-        C = ξ * w * ξ' / sum(w)
-    end
+    @assert size(wts) == batch_size(layer, x)
+    # reshape into a matrix even for unbatched `x`, where `flatten` gives a vector
+    ξ = reshape(flatten(layer, x .- mean), length(layer), :)
+    C = (ξ .* reshape(vec(wts), 1, :)) * ξ' / sum(wts)
```
cossio: can this use the three-arg `dot` function from LinearAlgebra?

cossio: Three-arg `dot(x, A, y)` computes the scalar `x'Ay`, but this line needs the full covariance matrix `ξ W ξ'`, so it doesn't apply directly. I switched the spelling to the standard LinearAlgebra three-factor product in 8e25417 — `ξ * Diagonal(vec(wts)) * ξ' / sum(wts)` — which is the form master used before this PR (same cost: the `Diagonal` scaling materializes one `ξ`-sized factor either way). Where a scalar `x'Wx` does occur (`initialize_w!`), it's already a two-arg `dot` of the weighted and unweighted factors.

---

src/train/infinite_minibatches.jl
```diff
@@ -54,51 +39,58 @@ function Base.iterate(iter::InfiniteMinibatchIterator, (i, shuffled))
...
+        wts::AbstractVector;
         batchsize::Int,
     )
     batchsize > 0 || throw(ArgumentError("batchsize must be positive"))
-    isnothing(wts) && return data, wts, nothing, batchsize
-
     length(wts) == size(data, ndims(data)) ||
         throw(DimensionMismatch("length(wts) must equal the number of data samples"))
+    wts = _validate_weights(wts)
+    data, wts = _filter_zero_weights(data, wts)
```
cossio: Don't filter zero-weight samples. That's the calling-user responsibility.

cossio: Done in 87943c7: `_filter_zero_weights` is gone and weights are used exactly as given. `_prepare_training_data` now only validates and returns `(mean(wts), clamped batchsize)`, the trainer `moments` default is eagerly `moments_from_samples(rbm.visible, data; wts)` again, and `initialize!` just validates. Zero-weight observations contribute zero to the weighted averages but keep their minibatch slots; docs and CHANGELOG now state that dropping them beforehand is the caller's responsibility. The filtering-dependent tests were removed and the affected suites plus the full suite are green locally.

---

src/util/util.jl
```diff
@@ -1,40 +1,59 @@
...
+[`wsum`](@ref)). By default, lazy uniform weights over all of `A`, which
+reduce like an ordinary `mean` without allocating a weights array or
+promoting eltypes.
+
+```math
+\frac{\sum_i A_i w_i}{\sum_i w_i}
+```
+"""
+function wmean(A::AbstractArray; wts::AbstractArray = Ones{Bool}(size(A)))
+    return wsum(A, wts) / sum(wts)
```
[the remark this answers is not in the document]
cossio: Declining per the repository owner's explicit direction on this PR to drop the machinery that guarded floating-point accuracy against extreme or narrow weights: reductions run in the weights' own precision, and `Float16` weights whose sum exceeds `floatmax(Float16) ≈ 65504` are outside the supported envelope, like the near-`floatmax` cases already documented. The CHANGELOG entry states that extreme weights can overflow and ordinary weights are unaffected; a caller who needs 70k `Float16` weights can convert them once at the call site.

---

src/layers/abstractlayer.jl
cossio: Done in 3bea81c, as agreed: the scalar `uniform_weights` method is gone, replaced by `batch_size(layer, ::Number) = ()` — a scalar input broadcasts over the layer and has no batch dimensions — so the generic `uniform_weights(layer, x) = Ones{Bool}(batch_size(layer, x))` now covers the scalar-inputs defaults through the ordinary path.

---

src/util/util.jl
```diff
@@ -1,40 +1,59 @@
...
+# kernel than float-converted copies, breaking exact reproducibility.
+_asfloat(A::AbstractArray{<:AbstractFloat}) = A
+_asfloat(A::AbstractArray) = float.(A)
+
+# `transpose`, not `dot`: the documented sum is `Σ Aᵢwᵢ`, without conjugation
+_wsum_all(A::AbstractArray, wts::AbstractArray) = transpose(_asfloat(vec(A))) * _asfloat(vec(wts))
+_wsum_trailing(A::AbstractMatrix, wts::AbstractVector) = _asfloat(A) * _asfloat(wts)
+# uniform `Ones` weights reduce as a plain sum (keeps the unweighted training
+# path free of weighted copies and eltype promotion)
+_wsum_trailing(A::AbstractMatrix, ::Ones{<:Real}) = sum(A; dims = 2)
```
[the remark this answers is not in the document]
cossio: Confirmed and fixed in 7e8ccf3: the trailing-dims `Ones` fast path now sums `_asfloat(A)`, so integer data accumulates in float (matching `Statistics.mean`, which promotes its accumulator) instead of wrapping in an integer `sum`. Verified `batchmean(Gaussian((1,)), reshape([typemax(Int), typemax(Int)], 1, 2)) ≈ 9.22e18`, and the integer/Bool moments-parity testset still passes. Regression tests added in `test/zero_weight_training.jl`.

---

src/util/util.jl
```diff
@@ -1,40 +1,59 @@
...
-_weighted_term(a, w) = iszero(w) ? zero(a) * w : a * w
+# Float the operands (a no-op for float arrays): narrow-integer data and
+# weights would otherwise accumulate in their own narrow type (wrapping on
+# overflow), and mixed integer/float operands would take a different matmul
+# kernel than float-converted copies, breaking exact reproducibility.
+_asfloat(A::AbstractArray{<:AbstractFloat}) = A
+_asfloat(A::AbstractArray) = float.(A)
+
+# `transpose`, not `dot`: the documented sum is `Σ Aᵢwᵢ`, without conjugation
+_wsum_all(A::AbstractArray, wts::AbstractArray) = transpose(_asfloat(vec(A))) * _asfloat(vec(wts))
```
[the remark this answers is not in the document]
cossio: Confirmed and fixed in 7e8ccf3, with a slightly different shape than suggested: instead of a second `Ones` reduction specialization, `_asfloat` now keeps `Ones{Bool}` weights as-is (they are exact ones; floating them to `Ones{Float64}` was pointless and promoted the matmul), so the existing `_wsum_all` covers the full uniform reduction without promotion. `wmean(Float32[1, 2])` now returns a `Float32`, verified along with eltype regression tests in `test/zero_weight_training.jl`.

---

src/train/infinite_minibatches.jl
```diff
@@ -54,51 +39,41 @@ function Base.iterate(iter::InfiniteMinibatchIterator, (i, shuffled))
...
-
     length(wts) == size(data, ndims(data)) ||
         throw(DimensionMismatch("length(wts) must equal the number of data samples"))
+    _validate_weights(wts)
+    return mean(wts), min(batchsize, length(wts))
+end
+
+# Real-valued lazy uniform weights are trivially valid; anything else
+# (including complex-valued `Ones`) goes through the elementwise checks.
+_validate_weights(wts::Ones{<:Real}) = wts
```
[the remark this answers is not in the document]
cossio: Confirmed and fixed in 7e8ccf3: the `Ones{<:Real}` fast path now rejects empty weights with the same `ArgumentError` as any weights lacking a positive entry, since empty weights have no positive weight whatever their type. `initialize!(rbm, zeros(len, 0))` and `_prepare_training_data` on empty data now fail before any mutation; regression tests added in `test/zero_weight_training.jl`.

---

src/train/train.jl
```diff
@@ -36,21 +36,24 @@ function _train!(
...
 
     for (iter, (vd, wd)) in zip(1:iters, infinite_minibatches(data, wts; batchsize, shuffle))
-        batch_weight = _batch_weight(wd, normalization)
+        batch_weight = _batch_weight(wd, wts_mean)
 
         # positive and negative phase gradients
         ∂d = ∂free_energy(rbm, vd; wts = wd, moments)
         ∂m, extras = negative_phase(vd)
-        ∂ = (∂d - ∂m) * batch_weight # correct weighted minibatch bias
+        # Correct the weighted minibatch bias, in the gradient eltype so the
```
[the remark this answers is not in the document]
cossio: The arithmetic is correct — a minibatch whose weights are all zero divides by `sum(wd) == 0`, and the zero `_batch_weight` cannot rescue it since `NaN * 0 = NaN`. But this is the deliberate, documented contract, not an oversight: the repository owner explicitly directed on this PR that zero-weight samples must not be filtered ("Don't filter zero-weight samples. That's the calling-user responsibility"), which is why the earlier filtering machinery was removed in 87943c7. Skipping zero-total-weight minibatches would re-introduce a per-iteration variant of exactly that machinery. Callers who keep zero weights sparse enough to fill a minibatch should drop those observations beforehand, as the docs, docstrings, and CHANGELOG state; c480517 makes the NaN consequence explicit in the training docs.

---

src/train/gradient.jl
```diff
@@ -68,20 +71,11 @@ function ∂interaction_energy(rbm::RBM, v::AbstractArray, h::AbstractArray; wts
...
-            @assert size(wts) == bsz
-            # Normalize like `wmean`, so extreme finite weights cannot
-            # overflow, and zero weights annihilate their samples exactly:
-            # both factors must be zeroed, else `NaN * 0` poisons the product.
-            wn = reshape(vec(wts) ./ (1.0 * float(maximum(wts))), 1, :)
-            vw = _weighted_term.(vflat, wn)
-            hz = ifelse.(iszero.(wn), zero.(hflat), hflat)
-            ∂wflat = -vw * hz' / sum(wn)
-        end
+        ∂wflat = -vflat * (hflat .* reshape(vec(wts), 1, :))' / sum(wts)
```
cossio: can't we use :

vflat  * Diagonal(vec(wts)) * hflat'
here ?

src/train/infinite_minibatches.jl
```diff
@@ -54,51 +39,58 @@ function Base.iterate(iter::InfiniteMinibatchIterator, (i, shuffled))
...
+    data, wts = _filter_zero_weights(data, wts)
+    return data, wts, mean(wts), min(batchsize, length(wts))
+end
+
+# Real-valued lazy uniform weights are trivially valid; anything else
+# (including complex-valued `Ones`) goes through the elementwise checks.
+_validate_weights(wts::Ones{<:Real}) = wts
+
+function _validate_weights(wts::AbstractArray)
     all(w -> w isa Real && isfinite(w) && w ≥ 0, wts) ||
```
cossio: `w ≥ 0` -> `w > 0`

Let's assert all weights are positive. That way zero weights error automatically.

src/train/infinite_minibatches.jl
```diff
@@ -54,51 +39,58 @@ function Base.iterate(iter::InfiniteMinibatchIterator, (i, shuffled))
...
         throw(ArgumentError("wts must contain only finite, nonnegative real values"))
-
-    positive = map(w -> !iszero(w), wts)
-    npositive = count(positive)
-    npositive > 0 ||
+    any(w -> !iszero(w), wts) ||
         throw(ArgumentError("wts must contain at least one positive weight"))
```
cossio: this is then not needed (by the previous comment)

src/train/infinite_minibatches.jl
```diff
@@ -54,51 +39,58 @@ function Base.iterate(iter::InfiniteMinibatchIterator, (i, shuffled))
...
+    return wts
+end
 
-    if npositive < length(wts)
-        # GPUArrays do not generally support logical indexing. Transfer only
-        # the mask used to build indices; indexing preserves the data backend.
-        positive_indices = findall(adapt(Array, positive))
-        data, wts = getobs(positive_indices, data, wts)
-    end
+_filter_zero_weights(data::AbstractArray, wts::Ones{<:Real}) = (data, wts)
```
cossio: remove this

src/train/infinite_minibatches.jl
```diff
@@ -54,51 +39,58 @@ function Base.iterate(iter::InfiniteMinibatchIterator, (i, shuffled))
...
+    positive = map(w -> !iszero(w), wts)
+    npositive = count(positive)
+    npositive == length(wts) && return data, wts
 
-    return data, wts, normalization, min(batchsize, npositive)
+    # GPUArrays do not generally support logical indexing. Transfer only
+    # the mask used to build indices; indexing preserves the data backend.
+    positive_indices = findall(adapt(Array, positive))
+    return getobs(positive_indices, data, wts)
 end
```
cossio: remove this

src/train/gradient.jl
cossio: Adopted in d10bf09, with one twist found while verifying on JLArrays: the spelling works for explicit weight vectors, but for the default lazy `Ones` weights, `Diagonal(Trues(n))` sends LinearAlgebra's `_mul_diag!` into its generic scalar-indexing loop when the factors are GPU arrays, which `allowscalar(false)` forbids (verified: unweighted `∂free_energy` on JLArrays fails with exactly this spelling; `batchcov` had the same latent issue). So both call sites now share a two-method helper in util.jl: `_weighted_outer(A, wts, B) = A * Diagonal(vec(wts)) * B'`, with `Ones` weights reducing to the plain `A * B'`. Weighted and unweighted paths verified on CPU and JLArrays.

---

src/train/infinite_minibatches.jl
cossio: Done in 382a69d: `_validate_weights` now requires `w > 0` (finite, positive reals), so zero weights raise `ArgumentError` automatically, before any mutation. Docstrings, docs, CHANGELOG, and tests updated to the positive-weights contract — observations meant to be excluded must be dropped (with their weights) before training. This also makes the zero-total-weight-minibatch NaN scenario impossible through the public trainers.

---

src/train/infinite_minibatches.jl
cossio: Removed in 382a69d — with `w > 0` asserted elementwise, the separate at-least-one-positive check is redundant.

---

src/train/infinite_minibatches.jl
cossio: This comment anchored on a stale version of the file — `_filter_zero_weights` (both methods) was already removed in 87943c7 ("Stop filtering zero-weight observations; caller's responsibility"), pushed last night; the current head has no filtering. Apologies for not closing the loop on the original thread when that landed. With 382a69d zero weights are now rejected outright.

---

src/train/infinite_minibatches.jl
cossio: Same as above — this method was already removed in 87943c7; the pushed head has no zero-weight filtering anywhere.

---

src/train/initialization.jl
```diff
@@ -140,16 +170,14 @@ Initializes `rbm.w` such that typical inputs to hidden units are λ.
...
-    else
-        @assert length(wts) == size(data)[end]
-        x = reshape(data, length(rbm.visible), size(data)[end])
-        d = dot(x * Diagonal(wts), x / sum(wts))
-    end
+    @assert length(wts) == size(data)[end]
+    x = reshape(data, length(rbm.visible), size(data)[end])
+    # dividing `x` first floats the accumulator, so narrow integer data
+    # (e.g. Int8 spins) cannot overflow the dot product
+    d = dot(x .* reshape(wts, 1, :), x / sum(wts))
```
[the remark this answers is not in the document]
cossio: Fixed in 4475302: both `dot` operands are now floated with `_asfloat` before the products, consistent with the other weighted reductions, so narrow integer data or weights (e.g. Int8 spins) can no longer wrap. A regression test checks that Int8 data with Int8 weights reproduces the float computation exactly.

---

src/train/infinite_minibatches.jl
```diff
@@ -54,51 +39,41 @@ function Base.iterate(iter::InfiniteMinibatchIterator, (i, shuffled))
...
+_validate_weights(wts::Ones{<:Real}) = wts
 
-# mean(wd) / mean(wts), overflow-safe: batch weights are a subset of the
-# training weights, so the global scale bounds them and its wide type
-# propagates through the broadcast.
-function _batch_weight(wd::AbstractVector, normalization::NamedTuple)
-    return mean(wd ./ normalization.scale) / normalization.mean
+function _validate_weights(wts::AbstractArray)
+    all(w -> w isa Real && isfinite(w) && w > 0, wts) ||
+        throw(ArgumentError("wts must contain only finite, positive real values"))
```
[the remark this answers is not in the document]
cossio: Fixed in 4475302, at the entry points rather than in `_validate_weights`: `_prepare_training_data` and RBM-level `initialize!` now raise `ArgumentError("data must contain at least one sample")` for empty data, before any mutation. This keeps the weight validation itself to the owner's minimal `all(w > 0)` form (vacuous truth on empty collections is standard Julia semantics — `mean(Float64[])` is `NaN` in the stdlib too) while making the entry points fail loudly. Tests cover both the trainer and `initialize!` paths.

---

src/train/initialization.jl
```diff
@@ -19,17 +19,26 @@ function initialize!(rbm::RBM; ϵ::Real = 1.0e-6)
...
     initialize!(rbm.hidden)
     initialize_w!(rbm, data; ϵ, wts)
     zerosum!(rbm)
     return rbm
 end
 
-function initialize!(layer::Binary, data::AbstractArray; ϵ::Real = 1.0e-6, wts = nothing)
+function initialize!(
+        layer::Binary, data::AbstractArray;
+        ϵ::Real = 1.0e-6, wts::AbstractArray = uniform_weights(layer, data)
```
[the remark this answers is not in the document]
cossio: Declining this one: the split is deliberate. Validation lives at the training entry points — `pcd!` (via `_prepare_training_data`) and RBM-level `initialize!` — while layer-level `initialize!` methods, like `wmean` and `∂free_energy`, are unvalidated building blocks; the owner trimmed validation to the minimal boundary form during review, and the tests document the two-tier contract explicitly. Repeating `_validate_weights` in each of the eight layer overloads is the kind of defensive scaffolding this PR has consistently removed. Happy to add it if the owner prefers layer-level checks.

---

src/layers/abstractlayer.jl
```diff
@@ -309,14 +323,18 @@ moment, the next axes are `size(layer)`, and any trailing batch dimensions of
...
-with batched inputs) over its batch dimensions, weighted by `wts`.
+with batched inputs) over its batch dimensions, weighted by `wts` (lazy uniform
+weights by default).
 """
-function batchmean_moments(layer::AbstractLayer, moments::AbstractArray; wts = nothing)
-    m = wmean(moments; wts, dims = (ndims(layer) + 2):ndims(moments))
-    return reshape(m, ntuple(d -> size(moments, d), Val(ndims(layer) + 1)))
+function batchmean_moments(
+        layer::AbstractLayer, moments::AbstractArray;
+        wts::AbstractArray = Ones{Bool}(size(moments)[(ndims(layer) + 2):end])
```
cossio, line 334: can this use:

```
uniform_weights(layer, view(moments, 1, size(moments)[2:end]...)
```

or something similar ?

src/train/infinite_minibatches.jl
```diff
@@ -54,51 +39,43 @@ function Base.iterate(iter::InfiniteMinibatchIterator, (i, shuffled))
...
     return InfiniteMinibatchIterator(ds, batchsize, shuffle)
 end
 
+#= Per-run weight checks. Weights must be finite, positive reals: zero
+weights are rejected, so observations meant to be excluded must be dropped
+(with their weights) before training. Weights are otherwise used as given —
+they are not rescaled (extreme finite weights can overflow the plain
+reductions). Returns the mean weight (used to bias-correct minibatch
+gradients) and the batchsize clamped to the number of samples. =#
 function _prepare_training_data(
```
cossio, line 53: Is this function really needed ?

src/train/infinite_minibatches.jl
```diff
@@ -54,51 +39,43 @@ function Base.iterate(iter::InfiniteMinibatchIterator, (i, shuffled))
...
+function _validate_weights(wts::AbstractArray)
+    all(w -> w isa Real && isfinite(w) && w > 0, wts) ||
+        throw(ArgumentError("wts must contain only finite, positive real values"))
+    return wts
 end
+
+# mean(wd) / mean(wts), the bias correction for a weighted minibatch: a
+# scalar in the weights' own precision, applied in the gradient eltype at its
+# use site.
+_batch_weight(wd::AbstractVector, wts_mean::Real) = mean(wd) / wts_mean
```
cossio, line 81: Is this function really needed ?

cossio: I think you can just inline this

cossio:
@claude /code-review:code-review --comment

src/layers/abstractlayer.jl
cossio: Done in a4c04e6, exactly as suggested: the default is now `uniform_weights(layer, view(moments, 1, ..))` — the first-moment view has the layer-then-batch shape `uniform_weights` expects, so the hand-computed size slicing is gone. The view is lazy (only its size is used) and the full suite, including JLArrays, passes.

---

src/train/infinite_minibatches.jl
cossio: No — removed in a4c04e6. `_prepare_training_data` had a single caller (`_train!`), so its checks and the two return values (`mean(wts)`, clamped batchsize) are now written inline there, before `setup!` mutates anything. Only `_validate_weights` remains as a function, since it is shared with `initialize!`. Tests of the removed helper now exercise `_validate_weights` and the public `pcd!` entry point directly.

---

src/train/infinite_minibatches.jl
cossio: Inlined in a4c04e6: `_batch_weight` is gone and `_train!` computes `batch_weight = mean(wd) / wts_mean` directly at its single use site.

---

src/layers/abstractlayer.jl
```diff
@@ -156,100 +156,111 @@ end
...
-        μ = wmean(x; wts, dims = batchdims(layer, x))
-        return reshape(μ, size(layer))
-    end
+    @assert size(wts) == batch_size(layer, x)
+    return wmean(x; wts)
 end
 
 """
-    batchvar(layer, x; wts = nothing, [mean])
+    batchvar(layer, x; wts = uniform_weights(layer, x), [mean])
```
cossio, line 189: Simpler to just write: 
> batchvar(layer, x; [wts, mean])
to indicate `wts,mean` are optional. Make the same in the other docstrings

cossio:
Closed in favor of #220, which carries the identical code (head `a4c04e6` plus a merge of current master) from a session that can monitor the PR. All review discussion remains here.

---

PR #214, opened 2026-08-08
Document the moments interface and dissolve layers/moments.jl

cossio:
Documents the per-layer moments interface and consolidates its code organization. No behavior change.

- Add generic stubs with docstrings for `moments_from_samples` and `∂energy_from_moments` in `abstractlayer.jl`, alongside the existing `moments_from_inputs` stub, stating the shared moments-array layout and the meaning of each moment slot per layer family.
- Dissolve `src/layers/moments.jl` into the files where its methods belong: `batchmean_moments` moves next to the moments interface stubs in `abstractlayer.jl`, the multi-family `mean_from_moments`/`var_from_moments` methods move to `common.jl`, and the Spin-specific variance method moves to `spin.jl`.
- Add documented stubs for `mean_from_moments` and `var_from_moments` in `abstractlayer.jl` (their methods are now spread across `common.jl` and `spin.jl`), so the whole moments interface is documented in one place.

All function bodies and method tables are unchanged.

https://claude.ai/code/session_01Mf1rQKnfZ7inH8gmpC7bo6


src/layers/abstractlayer.jl
```diff
@@ -258,6 +258,38 @@ function total_meanvar_from_inputs(layer::AbstractLayer, inputs = 0; wts = nothi
...
+
+Empirical moments of `data`: batch averages (weighted by `wts`) of the layer's
+sufficient statistics. The first axis indexes the moment and the remaining axes
+are `size(layer)` (batch dimensions of `data` are averaged over), so the result
+has the same shape as `layer.par`. The moment slots for each layer family are:
+
+- `Binary`, `Spin`, `Potts`, `PottsGumbel`: `<x>`.
+- `Gaussian`, `ReLU`: `<x>`, `<x^2>`.
+- `dReLU`, `pReLU`, `xReLU`, `nsReLU` (shared dReLU layout): `<xp>`, `<xn>`,
+  `<xp^2>`, `<xn^2>`, where `xp = max(x, 0)` and `xn = min(x, 0)`.
```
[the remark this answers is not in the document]
cossio: No change needed: every `<x>`-style token here is inside a backtick code span, and Markdown never parses HTML inside code spans — Julia's Markdown, Documenter, and GitHub all render them literally (Documenter emits `<code>&lt;x&gt;</code>`). Escaping would instead display the raw `&lt;`/`&gt;` entities in REPL help mode. This also matches the established notation elsewhere in the package, e.g. the `mean_from_moments` docstring ("Mean unit activations `<x>`"). These functions are not currently included in the Documenter manual in any case.

---

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file. The format
 
 ## Unreleased
 
+- Documented the moments interface: `moments_from_samples` and
+  `∂energy_from_moments` gained generic docstrings in `abstractlayer.jl`
+  (alongside the existing `moments_from_inputs` one) stating the shared
+  moments-array layout and the meaning of each moment slot per layer family.
+
```
cossio: this does not need to be in CHANGELOG.

cossio: Removed in 9ce58dc.

---

src/layers/abstractlayer.jl
cossio: This does not go here. It goes in the doctoring of each specific layer method

cossio: Here, leave a general comment saying that each layer can define which moments it computes , that generally they are sufficient statistics which do not depend on parameters (on the lines of the conversation we had) ... but keep it succinct.

cossio: Done in 4a5eaca + 4e20186: the per-family slot lists moved to docstrings on each specific `moments_from_samples` method (`common.jl`, `gaussian.jl`, `relu.jl`), and the generic docstring now states succinctly that each layer defines which moments it computes — generally the sufficient statistics of the layer distribution, independent of the layer parameters, hence computable once per dataset and reused during training.

---

cossio:
The review finding above (docstring shape claim false for `nsReLU`) was confirmed and is fixed in follow-up PR #216, since this PR merged before the fix could land here.

---

PR #217, opened 2026-08-08
Use false as the zero in dReLU-family sample clipping

cossio:
Zero-clipping with an `Int` literal (`max.(x, 0)` / `min.(x, 0)`) widens narrow-integer or `Bool` arrays to `Int64` through promotion. `false` is a neutral zero under Julia's promotion rules (`promote_type(T, Bool) == T` for all numeric `T`), so clipping against `false` is value-identical while preserving the array eltype. Float data is unaffected in both value and type.

Changes:
- `moments_from_samples` for the dReLU-family layers (`dReLU`, `pReLU`, `xReLU`, `nsReLU`) clips with `max.(data, false)` / `min.(data, false)`. This is the one zero-clip site that operates on user-supplied data, so it is the only place where the `Bool` zero changes anything; the other clip sites (`xReLU -> dReLU` conversion, `mode_from_inputs(::ReLU, ...)`) operate on quantities that are already float and keep the plain `0`.
- The second moments are computed as `abs2.(float.(x))` rather than `x .^ 2`: with the eltype now preserved, squaring narrow integers in their own type could overflow (`Int8(127)^2` wraps to `1`), so the squares go through `float`, which is an eltype no-op for float samples. (Found by Codex review.)
- New testset checks the moments of `Int8` (including extreme values `±127`/`-128`) and `Bool` data, batched, unbatched, and weighted, against those of the float-converted samples.

No changelog entry: observable API results for float samples are unchanged; for integer samples only intermediate eltypes and overflow behavior at unrealistic magnitudes differ.

Tests: `test/layers.jl` passes (48 testsets) on top of the merged master.


src/layers/common.jl
```diff
@@ -118,8 +118,8 @@ pReLU(layer::Gaussian) = pReLU(dReLU(layer))
 xReLU(layer::Gaussian) = xReLU(dReLU(layer))
 
 function moments_from_samples(layer::Union{dReLU, pReLU, xReLU, nsReLU}, data::AbstractArray; wts = nothing)
-    xp = max.(data, 0)
-    xn = min.(data, 0)
+    xp = max.(data, false)
+    xn = min.(data, false)
```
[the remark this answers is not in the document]
cossio: Added in 547c712: the new testset `"moments_from_samples with integer and Bool data"` exercises all four dReLU-family layers with `Int8` data (including the extreme values `127`/`-128`) and `Bool` data, batched, unbatched, and weighted. It asserts the moments equal those of the float-converted samples rather than asserting intermediate eltypes: the clipped intermediates aren't observable through the API (`batchmean` promotes when averaging, and the second moments now deliberately go through `float` to avoid narrow-integer overflow), and checking `eltype(max.(data, false))` directly would test Julia's promotion rules rather than this package. The value equality at `Int8` extremes is the observable regression guard.

---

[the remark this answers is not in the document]
cossio: Good catch — confirmed and fixed in 547c712. The second moments are now computed as `abs2.(float.(xp))` / `abs2.(float.(xn))`, so narrow integers are widened before squaring while float samples keep their eltype (`float` is an elementwise no-op for them, and the fused broadcast allocates the same single array `xp .^ 2` did). The first moments still clip without widening, which is safe because `mean`/`wmean` promote their accumulator. A regression testset now checks the moments of `Int8` data including `127` and `-128` against the float-converted samples, which fails under the wraparound you describe.

---

src/layers/common.jl
```diff
@@ -98,8 +98,8 @@ function xReLU(layer::dReLU)
 end
 
 function dReLU(layer::xReLU)
-    ξp = @. (1 + abs(layer.ξ)) / (1 + max(2layer.ξ, 0))
-    ξn = @. (1 + abs(layer.ξ)) / (1 - min(2layer.ξ, 0))
+    ξp = @. (1 + abs(layer.ξ)) / (1 + max(2layer.ξ, false))
+    ξn = @. (1 + abs(layer.ξ)) / (1 - min(2layer.ξ, false))
```
cossio: Here it's fine to use `0`. There is already a multiplication by an `Int`, so `false` here buys nothing

src/layers/relu.jl
```diff
@@ -22,7 +22,7 @@ end
 
 cgfs(layer::ReLU, inputs = 0) = relu_cgf.(layer.θ .+ inputs, layer.γ)
 sample_from_inputs(layer::ReLU, inputs = 0) = relu_rand.(layer.θ .+ inputs, layer.γ)
-mode_from_inputs(layer::ReLU, inputs = 0) = max.((layer.θ .+ inputs) ./ abs.(layer.γ), 0)
+mode_from_inputs(layer::ReLU, inputs = 0) = max.((layer.θ .+ inputs) ./ abs.(layer.γ), false)
```
cossio: fine to use `0` (same as in previous comment)

src/layers/common.jl
cossio: Agreed — reverted to `0` in 0eeb3c0. `false` now appears only in `moments_from_samples`, where the clipped array is user-supplied data and the eltype preservation actually does something.

---

src/layers/relu.jl
cossio: Reverted to `0` in 0eeb3c0 — the division already makes the argument float here, so `false` bought nothing.

---

PR #220, opened 2026-08-09
Unify training and statistics on a single weighted code path

cossio:
Supersedes #212 (closed so this session can own and monitor the PR — same code, full review history in #212's threads). Includes merges of master through #221 (`inputs::AbstractArray` with lazy `Falses` defaults, allowing removal of the scalar `batch_size` method) and #224 (explicit PCD loops, with the weighted path ported into each loop).

## Summary

Every previously unweighted code path is replaced by the weighted one, with `wts` defaulting to lazy uniform weights (`FillArrays.Trues`) — no `isnothing`/`isa` control flow and no `wts = nothing` anywhere. Incorporates #218 (closed in favor of this PR): internal weight rescaling is removed entirely. StatsBase was evaluated as a replacement for `wmean` and rejected: its weighted mean fails on GPU arrays (scalar indexing) and under Zygote (internal mutation), both verified; the stdlib matmul composition used here is the ecosystem-standard GPU/AD-safe route and needs no dependency.

- **`wmean(A; wts) = wsum(A, wts) / sum(wts)`**: weights act on the batch dimensions, which are always the trailing dimensions of `A`, so the reduced dimensions are inferred from `size(wts)` (no `dims` keyword; reduced dimensions are dropped). `wsum` splits by dispatch: a same-`ndims` method for full reductions and a trailing-reduction method, each with a uniform-`Trues` counterpart that reduces as a plain `sum(float, A[; dims])` — no-copy, float-accumulating, device-safe — and `sum(Trues(n)) = n::Int`, so `wmean` degenerates to `mean` exactly, without promoting eltypes.
- **`batchmean` always routes through `wmean`** — no unbatched special case; weights always multiply and divide. Unbatched inputs (including the lazy `Falses(size(layer))` defaults from #221) flow through the same path with 0-dimensional `Trues`.
- **Weighted outer products (`∂w` gradient, `batchcov`) are Diagonal-weighted matmuls**: `_weighted_outer(A, wts, B) = A * Diagonal(vec(wts)) * B'`, with lazy `Trues` weights reducing to the plain product (a `Diagonal` around a lazy CPU fill takes a scalar-indexing kernel on GPU arrays, verified with JLArrays under `allowscalar(false)`).
- Integer/Bool **data** is floated in the weighted matmuls (`float`, identity on float arrays) so it reduces bit-identically to its float-converted copy (required by the integer/Bool moments testset from #217; the mixed-eltype kernel differs from BLAS at the last ulp). Weights are used as given: they must have concrete real eltypes — `wts::AbstractArray{<:Real}` throughout, abstract-eltype arrays unsupported.
- `wts` defaults to lazy uniform `Trues` via `uniform_weights` **everywhere** — statistics kernels, `initialize!` (RBM-level, per-layer, and `initialize_w!`, all validating weights), and the three `pcd!` trainers. **Breaking**: `wts = nothing` is no longer accepted.
- **Weights must be finite, positive reals**, validated once per run before any mutation; zero or negative weights raise `ArgumentError`, so observations meant to be excluded must be dropped (with their weights) beforehand. Non-real weights are rejected by the signatures. Valid weights are used exactly as given: never rescaled (adopted from #218). Extreme weights (near `floatmax`, or needing wider-than-`Float64` accumulation) can overflow the plain reductions; ordinary weights are unaffected.

## Behavioral changes

1. **Breaking**: `pcd!` and `initialize!` require `wts::AbstractVector{<:Real}` (lazy uniform default); `nothing` is rejected.
2. **Breaking**: weights must be strictly positive — zero, negative, or non-finite weights raise `ArgumentError` before any computation touches the model, at every entry point (`pcd!` trainers, RBM-level and per-layer `initialize!`, `initialize_w!`); non-real weights fail at the signatures. Kernels (`wmean`, `∂free_energy`) remain plain unvalidated weighted reductions.
3. **Callbacks** receive lazy `Trues` slices as `wd` for unweighted training (previously `nothing`); weighted-training callbacks receive the raw minibatch weights, as before.
4. Unweighted `batchsize > nsamples` now clamps the batchsize instead of silently performing zero training iterations.
5. The extreme-weight scale-stability guarantee and its tests are removed.

Also fixed in passing: `batchcov` on unbatched data.

## Tests

Updated for the new contracts (`test/util.jl`, `test/zero_weight_training.jl`, `test/infinite_minibatches.jl`, `test/layers.jl`, `test/prelu_validation.jl`, `test/standardized.jl`, `test/jlarrays.jl`), including device regression tests for the uniform full reduction. Full suite green locally on Julia 1.12.6, including JLArrays with `allowscalar(false)`.

## Version

Contains breaking API changes (`wts = nothing` rejected, zero weights rejected by public trainers); suggest **7.0.0** per ColPrac — final call is the owner's, `Project.toml` untouched.

https://claude.ai/code/session_01KW4Gbdu8AvoTaTb9qabDtG

---
_Generated by [Claude Code](https://claude.ai/code/session_01KW4Gbdu8AvoTaTb9qabDtG)_


src/layers/abstractlayer.jl
```diff
@@ -156,100 +156,111 @@ end
...
+"""
+    batchmean(layer, x; wts = uniform_weights(layer, x))
 
 Mean of `x` over batch dimensions, weigthed by `wts`.
 """
-function batchmean(layer::AbstractLayer, x::AbstractArray; wts = nothing)
+function batchmean(
+        layer::AbstractLayer, x::AbstractArray; wts::AbstractArray = uniform_weights(layer, x)
+    )
     @assert size(layer) == size(x)[1:ndims(layer)]
```
cossio: this is already checked by `batch_size` just below, right ? If so, remove it here.

src/layers/abstractlayer.jl
```diff
@@ -156,100 +156,111 @@ end
...
+    batchvar(layer, x; wts = uniform_weights(layer, x), [mean])
 
 Variance of `x` over batch dimensions, weigthed by `wts`.
 """
 function batchvar(
-        layer::AbstractLayer, x::AbstractArray; wts = nothing, mean = batchmean(layer, x; wts)
+        layer::AbstractLayer, x::AbstractArray;
+        wts::AbstractArray = uniform_weights(layer, x), mean = batchmean(layer, x; wts)
     )
     @assert size(layer) == size(x)[1:ndims(layer)] == size(mean)
```
cossio: this is already checked by batch_size just below, right ? If so, remove it here.

Fix this whenever it happens throughout this PR

cossio: I meant, it's already checked in the line below, in batch_mean

src/layers/abstractlayer.jl
```diff
@@ -156,100 +156,111 @@ end
...
 """
-    batchvar(layer, x; wts = nothing, [mean])
+    batchvar(layer, x; wts = uniform_weights(layer, x), [mean])
 
 Variance of `x` over batch dimensions, weigthed by `wts`.
 """
 function batchvar(
-        layer::AbstractLayer, x::AbstractArray; wts = nothing, mean = batchmean(layer, x; wts)
+        layer::AbstractLayer, x::AbstractArray;
+        wts::AbstractArray = uniform_weights(layer, x), mean = batchmean(layer, x; wts)
```
cossio: mean::AbstractArray

src/layers/abstractlayer.jl
```diff
@@ -156,100 +156,111 @@ end
...
 """
-    batchstd(layer, x; wts = nothing, [mean])
+    batchstd(layer, x; wts = uniform_weights(layer, x), [mean])
 
 Standard deviation of `x` over batch dimensions, weigthed by `wts`.
 """
 function batchstd(
-        layer::AbstractLayer, x::AbstractArray; wts = nothing, mean = batchmean(layer, x; wts)
+        layer::AbstractLayer, x::AbstractArray;
+        wts::AbstractArray = uniform_weights(layer, x), mean = batchmean(layer, x; wts)
```
cossio: mean::AbstractArray

src/layers/abstractlayer.jl
```diff
@@ -156,100 +156,111 @@ end
...
 """
-    batchcov(layer, x; wts = nothing, [mean])
+    batchcov(layer, x; wts = uniform_weights(layer, x), [mean])
 
 Covariance of `x` over batch dimensions, weigthed by `wts`.
 """
 function batchcov(
-        layer::AbstractLayer, x::AbstractArray; wts = nothing, mean = batchmean(layer, x; wts)
+        layer::AbstractLayer, x::AbstractArray;
+        wts::AbstractArray = uniform_weights(layer, x), mean = batchmean(layer, x; wts)
```
cossio: mean::AbstractArray

src/layers/abstractlayer.jl
```diff
@@ -309,14 +323,18 @@ moment, the next axes are `size(layer)`, and any trailing batch dimensions of
...
-function batchmean_moments(layer::AbstractLayer, moments::AbstractArray; wts = nothing)
-    m = wmean(moments; wts, dims = (ndims(layer) + 2):ndims(moments))
-    return reshape(m, ntuple(d -> size(moments, d), Val(ndims(layer) + 1)))
+function batchmean_moments(
+        layer::AbstractLayer, moments::AbstractArray;
+        wts::AbstractArray = uniform_weights(layer, view(moments, 1, ..))
+    )
+    @assert ndims(wts) == ndims(moments) - ndims(layer) - 1
+    return wmean(moments; wts)
 end
```
cossio, line 338: Are we sure this is correct ?

src/train/gradient.jl
```diff
@@ -52,10 +52,13 @@ function ∂free_energy_h(
     return ∂RBM(∂v, ∂h, ∂w)
 end
 
-function ∂interaction_energy(rbm::RBM, v::AbstractArray, h::AbstractArray; wts = nothing)
+function ∂interaction_energy(
+        rbm::RBM, v::AbstractArray, h::AbstractArray;
+        wts::AbstractArray = Ones{Bool}(batch_size(rbm, v, h))
```
cossio: `Ones{Bool}` -> `Trues`

Everywhere

src/layers/abstractlayer.jl
cossio: Right — the `size(wts) == batch_size(layer, x)` assert on the next line calls `batch_size`, which performs exactly this check. Removed in a30e405.

---

src/layers/abstractlayer.jl
cossio: Done in a30e405, swept across the PR: removed here in `batchvar` and in `batchcov` (the `batchmean`/`flatten` calls below re-check the layer/data shapes), and in the five `initialize!` sites (Binary, Spin, Potts, the shared Gaussian helper, nsReLU) where `batchmean` performs the same check. One nuance: these asserts also pinned `size(mean) == size(layer)`, which nothing downstream fully enforces (a broadcast-compatible wrong-shape `mean`, e.g. a trailing singleton, expands silently); the new `mean::AbstractArray` annotation rejects scalars by dispatch, and the fully-wrong shapes still error in the broadcast.

---

src/train/initialization.jl
```diff
@@ -19,17 +19,28 @@ function initialize!(rbm::RBM; ϵ::Real = 1.0e-6)
...
     initialize!(rbm.hidden)
     initialize_w!(rbm, data; ϵ, wts)
     zerosum!(rbm)
     return rbm
 end
 
-function initialize!(layer::Binary, data::AbstractArray; ϵ::Real = 1.0e-6, wts = nothing)
+function initialize!(
+        layer::Binary, data::AbstractArray;
+        ϵ::Real = 1.0e-6, wts::AbstractArray = uniform_weights(layer, data)
```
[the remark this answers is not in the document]
cossio: Fixed in a30e405: every data-taking layer `initialize!` (Binary, Spin, Potts directly; the Gaussian/ReLU/dReLU/pReLU/xReLU family via the shared `_initialize_gaussian_moments!` helper; nsReLU) and `initialize_w!` now run `_validate_weights(wts)` before the weighted statistics, matching the RBM-level entry point, so zero/negative/non-finite weights raise the documented `ArgumentError`. Layer-level and `initialize_w!` cases added to the "initialize! weight checks" testset.

---

src/util/util.jl
```diff
@@ -1,40 +1,70 @@
...
+_wsum_trailing(A::AbstractMatrix, wts::AbstractVector) = _asfloat(A) * _asfloat(wts)
+# uniform `Ones` weights reduce as a plain sum (keeps the unweighted training
+# path free of weighted copies and eltype promotion)
+_wsum_trailing(A::AbstractMatrix, ::Ones{<:Real}) = sum(_asfloat(A); dims = 2)
+
+# `A * Diagonal(vec(wts)) * B'`, the weighted outer product `Σᵢ wᵢ A[:,i] B[:,i]'`.
+# Uniform `Ones` weights reduce to the plain product: `Diagonal` of a lazy CPU
+# fill takes a scalar-indexing kernel when the factors are GPU arrays.
+_weighted_outer(A::AbstractMatrix, wts::AbstractArray, B::AbstractMatrix) =
+    A * Diagonal(vec(wts)) * B'
```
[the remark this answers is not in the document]
cossio: Fixed in a30e405: `_weighted_outer` now builds `Diagonal(_asfloat(vec(wts)))`, the same narrowing/float path as `wsum`, so `Real[]`/`Any[]` weights produce a concretely typed gradient (verified the `eltype == Any` failure before the fix). Test added asserting `Float64` eltype and value equality with concretely typed weights.

---

src/util/util.jl
```diff
@@ -1,40 +1,70 @@
...
+# overflow), and mixed integer/float operands would take a different matmul
+# kernel than float-converted copies, breaking exact reproducibility.
+_asfloat(A::AbstractArray{<:AbstractFloat}) = A
+_asfloat(A::AbstractArray) = float.(A)
+# The default lazy uniform weights are exact ones already; keep them lazy (and
+# `Bool`) so they cannot promote the eltype of the reduction.
+_asfloat(A::Ones{Bool}) = A
+
+# `transpose`, not `dot`: the documented sum is `Σ Aᵢwᵢ`, without conjugation
+_wsum_all(A::AbstractArray, wts::AbstractArray) = transpose(_asfloat(vec(A))) * _asfloat(vec(wts))
```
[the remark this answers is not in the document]
cossio: Fixed in a30e405 — verified the scalar-indexing failure with JLArrays under `allowscalar(false)` before fixing. `_wsum_all` gets the uniform fast path `sum(_asfloat(A))`, so a full reduction with default weights is a plain device `sum`. Note this reinstates the `Ones` method that was removed earlier in review as a no-benefit optimization — it returns for device correctness, not speed (the code comment says so), and a JLArrays regression test now pins it.

---

src/layers/abstractlayer.jl
cossio: Done in a30e405, in `batchvar`, `batchstd`, and `batchcov`.

---

src/layers/abstractlayer.jl
cossio: Done in a30e405.

---

src/layers/abstractlayer.jl
cossio: Done in a30e405.

---

src/layers/abstractlayer.jl
cossio: Yes — the worry is the right thing to check, but `batchmean_moments` never assumes the moments' first-axis length matches `layer.par`. Every `moments_from_*` implementation returns `stack([...]; dims = 1)`, so `moments` always has the canonical layout `(M, size(layer)..., batch...)` with `M` free (1 for field layers, 2 for Gaussian/ReLU, 4 for the dReLU family; `M ≠ size(par, 1)` is fine — Potts has `par` of size `(1, Q, N...)` and this function never touches `par`). The default weights peel off the moment axis with `view(moments, 1, ..)` (index 1 exists whenever `M ≥ 1`) purely to read the batch dims via `batch_size`, which also verifies `size(layer)` against the middle axes. The assert's `- 1` is that moment axis, independent of `M` — contrast `∂cgf`, which asserts against `ndims(layer.par)` because there the first axis *is* the parameter axis. `wmean` then reduces exactly the trailing batch dims (`wsum` shape-checks `wts` against them), returning `(M, size(layer)...)`, whose `par`-shaped prefix `∂energy_from_moments` re-checks on its side. The unbatched case degenerates correctly too: 0-dimensional `Trues`, assert `0 == 0`, identity mean.

---

src/train/gradient.jl
cossio: Done in a30e405, across src and tests (imports updated; `Trues === Ones{Bool}`, so no behavior change). Plain `Ones` remains only where it isn't the `Bool` uniform-weight case: the `Ones{<:Real}` dispatch signatures in `util.jl`, the `Ones{eltype(w)}` scale fills in `offset_rbms.jl`/`standardized.jl`, and one `Ones{ComplexF64}` rejection test.

---

src/layers/abstractlayer.jl
```diff
@@ -156,100 +156,111 @@ end
...
     @assert size(layer) == size(x)[1:ndims(layer)]
     return size(x)[batchdims(layer, x)]
 end
+batch_size(::AbstractLayer, ::Number) = ()
 
 """
-    batchmean(layer, x; wts = nothing)
+    uniform_weights(layer, x)
+
+Lazy uniform weights (`FillArrays.Ones`) over the batch dimensions of `x`.
```
cossio: Replace with:

Lazy uniform weights over the batch dimensions of `x`.

The  (`FillArrays.Ones`) is implementation detail, no need to put it in docstrings.

Do the same in similar docstrings throughout this PR

cossio: Done in a3f67c1, swept: the `uniform_weights` docstring reads as you wrote it, the `pcd!` keyword list drops the `=Trues(nsamples)` default (prose already says "lazy uniform weights by default"), and the `wmean` docstring signature is `wmean(A; wts)` with the default described in prose. Remaining `Trues`/`Ones` mentions are in code and code comments only.

---

cossio:
The `_asfloat` copy on the uniform fast paths is fixed in 53b7b49: both `Ones` methods now use `sum(float, A)` / `sum(float, A; dims = 2)`, which accumulate non-float data in float per element without materializing a converted copy (verified bit-identical to the copy-based reduction across Bool/Int8/Int/Float32/Float64, ~1 KB vs ~8 MB allocated on a 100×10000 Bool batch, and a plain device mapreduce with JLArrays under `allowscalar(false)`). `_asfloat` remains on the matmul-based weighted paths, which need materialized float operands. `test/util.jl`, `test/layers.jl` (the integer/Bool moments bit-identity testsets), and `test/jlarrays.jl` pass with the change.

---

src/layers/abstractlayer.jl
```diff
@@ -164,92 +164,104 @@ function batch_size(layer::AbstractLayer, x::AbstractArray)
 end
 
 """
-    batchmean(layer, x; wts = nothing)
+    uniform_weights(layer, x)
+
+Lazy uniform weights over the batch dimensions of `x`.
+"""
+uniform_weights(layer::AbstractLayer, x) = Trues(batch_size(layer, x))
```
cossio: can we enforce x::AbstractArray in the args here ?

src/layers/abstractlayer.jl
```diff
@@ -164,92 +164,104 @@ function batch_size(layer::AbstractLayer, x::AbstractArray)
...
-    ξ = flatten(layer, x .- mean)
-    if isnothing(wts)
-        C = ξ * ξ' / size(ξ, 2)
-    else
-        @assert size(wts) == batch_size(layer, x)
-        w = Diagonal(vec(wts))
-        C = ξ * w * ξ' / sum(w)
-    end
+    @assert size(wts) == batch_size(layer, x)
+    # reshape into a matrix even for unbatched `x`, where `flatten` gives a vector
```
cossio: comment not needed

src/layers/abstractlayer.jl
cossio: Done in d512d0b. The untyped `x` existed only for the scalar-`inputs` case, which #221 removed from the contract — with the scalar `batch_size` method gone (8b35d3b) the annotation is now exact.

---

src/layers/abstractlayer.jl
cossio: Removed in b0522ec.

---

src/layers/abstractlayer.jl
```diff
@@ -164,92 +164,104 @@ function batch_size(layer::AbstractLayer, x::AbstractArray)
...
+    ξ = reshape(flatten(layer, x .- mean), length(layer), :)
+    C = _weighted_outer(ξ, wts, ξ) / sum(wts)
     return reshape(C, size(layer)..., size(layer)...)
 end
 
 """
-    total_mean_from_inputs(layer, inputs; wts = nothing)
+    total_mean_from_inputs(layer, [inputs]; wts = uniform_weights(layer, inputs))
```
cossio: total_mean_from_inputs(layer, [inputs]; [wts])

same for similar docstrings throughout this PR

cossio: Done in fe6c8b0, swept across the PR: the three `total_*_from_inputs`, `∂cgf`, `batchmean`/`batchvar`/`batchstd`/`batchcov`, `moments_from_samples` (generic and the four layer-specific docstrings), `∂energy`, `batchmean_moments`, `wmean`, and the six `total_*_h_from_v`/`total_*_v_from_h` wrappers in `rbm.jl` now all use `[wts]` (and `[mean]` where applicable) instead of spelling the defaults.

---

src/layers/abstractlayer.jl
```diff
@@ -309,14 +323,18 @@ moment, the next axes are `size(layer)`, and any trailing batch dimensions of
...
+weights by default).
 """
-function batchmean_moments(layer::AbstractLayer, moments::AbstractArray; wts = nothing)
-    m = wmean(moments; wts, dims = (ndims(layer) + 2):ndims(moments))
-    return reshape(m, ntuple(d -> size(moments, d), Val(ndims(layer) + 1)))
+function batchmean_moments(
+        layer::AbstractLayer, moments::AbstractArray;
+        wts::AbstractArray = uniform_weights(layer, view(moments, 1, ..))
+    )
+    @assert ndims(wts) == ndims(moments) - ndims(layer) - 1
```
cossio: Can we check sizes, not just ndims ?

cossio: Done in efb1666: `@assert size(wts) == batch_size(layer, view(moments, 1, ..))` — checks the exact batch shape and (via `batch_size`) that the middle axes match `size(layer)`, which the old ndims form ignored. Applied the same to the analogous ndims-only assert in `∂cgf`.

---

src/train/train.jl
```diff
@@ -36,21 +36,33 @@ function _train!(
...
 
-    data, wts, normalization, batchsize = _prepare_training_data(data, wts; batchsize)
+    batchsize > 0 || throw(ArgumentError("batchsize must be positive"))
+    size(data, ndims(data)) > 0 ||
+        throw(ArgumentError("data must contain at least one sample"))
+    length(wts) == size(data, ndims(data)) ||
+        throw(DimensionMismatch("length(wts) must equal the number of data samples"))
+    _validate_weights(wts)
+    wts_mean = mean(wts)
+    batchsize = min(batchsize, length(wts)) # a too-large batchsize clamps to one full batch
```
cossio: comment not needed

cossio: Removed in a4bd7e1.

---

src/train/initialization.jl
```diff
@@ -19,71 +19,100 @@ function initialize!(rbm::RBM; ϵ::Real = 1.0e-6)
...
+function initialize!(
+        rbm::RBM, data::AbstractArray;
+        ϵ::Real = 1.0e-6, wts::AbstractVector = Trues(size(data, ndims(data)))
+    )
     @assert 0 < ϵ < 1 / 2
     @assert size(data) == (size(rbm.visible)..., size(data)[end])
+    size(data, ndims(data)) > 0 ||
+        throw(ArgumentError("data must contain at least one sample"))
+    length(wts) == size(data, ndims(data)) ||
+        throw(DimensionMismatch("length(wts) must equal the number of data samples"))
```
cossio: merge into one check:

```
@assert length(wts) == size(data, ndims(data)) > 0
```

cossio: Done in 44cb3eb, exactly as written; the two tests pinning the old `ArgumentError`/`DimensionMismatch` types now expect `AssertionError`. Note `pcd!` (`_train!`) still uses the throw-based versions of these same checks — say the word if you want the assert chain there too.

---

src/train/pcd.jl
```diff
@@ -44,7 +45,7 @@ function pcd!(
         data::AbstractArray;
         batchsize::Int = 1,
         iters::Int = 1, # number of gradient updates
-        wts::Union{AbstractVector, Nothing} = nothing, # data weights
+        wts::AbstractVector = Trues(size(data, ndims(data))), # data weights
```
cossio: can't we use `uniform_weights` here ?

src/train/initialization.jl
```diff
@@ -140,16 +172,15 @@ Initializes `rbm.w` such that typical inputs to hidden units are λ.
 """
 function initialize_w!(
         rbm::RBM, data::AbstractArray;
-        λ::Real = 0.1, ϵ::Real = 1.0e-6, wts::Union{Nothing, AbstractVector} = nothing
+        λ::Real = 0.1, ϵ::Real = 1.0e-6, wts::AbstractVector = Trues(size(data, ndims(data)))
```
cossio: can't we use uniform_weights here ?

src/train/pcd.jl
cossio: Done in 7ab676b, at all five `Trues(size(data, ndims(data)))` sites: the three `pcd!` trainers, RBM-level `initialize!`, and `initialize_w!` now default `wts = uniform_weights(rbm.visible, data)`.

---

src/train/initialization.jl
cossio: Done in 7ab676b (same sweep as the `pcd!` thread — all five sites).

---

src/RestrictedBoltzmannMachines.jl
```diff
@@ -2,9 +2,9 @@ module RestrictedBoltzmannMachines
 
 import ChainRulesCore
 import LinearAlgebra
-using Adapt: Adapt, adapt
```
cossio, line 5: > using Adapt: Adapt

Can we just remove this line ?

cossio: Not without losing device transfer: `src/adapt.jl` uses the `Adapt` name for its `Adapt.@adapt_structure` declarations on every layer type and RBM struct — that's what makes `adapt(CuArray, rbm)` / `adapt(JLArray, rbm)` recurse through the structs (exercised by the "adapt round-trip" JLArrays testset). This PR did trim the line from `using Adapt: Adapt, adapt` to just the module name, since the rewrite removed the last direct `adapt(...)` call in `src/`; the remaining import is the minimum `src/adapt.jl` needs.

---

src/train/infinite_minibatches.jl
```diff
@@ -54,51 +39,21 @@ function Base.iterate(iter::InfiniteMinibatchIterator, (i, shuffled))
...
+(including complex-valued `Ones`) goes through the elementwise checks. =#
+_validate_weights(wts::Ones{<:Real}) = wts
 
-# mean(wd) / mean(wts), overflow-safe: batch weights are a subset of the
-# training weights, so the global scale bounds them and its wide type
-# propagates through the broadcast.
-function _batch_weight(wd::AbstractVector, normalization::NamedTuple)
-    return mean(wd ./ normalization.scale) / normalization.mean
+function _validate_weights(wts::AbstractArray)
+    all(w -> w isa Real && isfinite(w) && w > 0, wts) ||
```
cossio: > w isa Real

instead of this, better to do

wts::AbstractArray{<:Real}

everywhere

cossio: Done in 2330958: all 49 `wts` signatures across src are now `AbstractArray{<:Real}`/`AbstractVector{<:Real}`, and `_validate_weights` drops the elementwise `w isa Real` (keeping only finite/positive). Non-real weights now fail at the signature — `TypeError` for keyword annotations (e.g. `pcd!(...; wts = ComplexF64[...])`), `MethodError` for positional dispatch — instead of `ArgumentError`; tests and CHANGELOG updated accordingly. One consequence: `Any[...]`-eltype weights are now rejected outright rather than validated elementwise (`Real[...]` still works).

Also in f21e996, from the chat discussion: `_wsum_all`/`_wsum_trailing` are gone in favor of a top-level `wsum(A, wts::Ones{<:Real})` dispatch method, with the generic `wsum` absorbing the matmul expressions.

---

src/util/util.jl
```diff
@@ -1,40 +1,75 @@
...
-        @assert size(wts) == size(A)
-        w = wts
+The reduced dimensions are inferred from the shape of `wts`, which must match
+the trailing dimensions of `A` (all of `A` for a full reduction, returning a
+scalar). Reduced dimensions are dropped from the result.
+"""
+function wsum(A::AbstractArray, wts::AbstractArray{<:Real})
+    kept = ndims(A) - ndims(wts)
+    @assert kept ≥ 0
+    @assert size(wts) == ntuple(i -> size(A, kept + i), ndims(wts))
```
cossio: @assert size(wts) == size(A)[ndims(A)-ndims(wts)+1:end]

src/util/util.jl
```diff
@@ -1,40 +1,75 @@
...
-    wn = w ./ (1.0 * float(maximum(wts)))
-    return mean(_weighted_term.(A, wn); dims) ./ mean(wn)
+# Uniform weights reduce as a plain sum: `float` per element accumulates
+# non-float data in float without materializing a converted copy, and the
+# lazy CPU fill would take a scalar-indexing kernel in the mixed matmul
+# when `A` is a GPU array.
+function wsum(A::AbstractArray, wts::Ones{<:Real})
+    kept = ndims(A) - ndims(wts)
+    @assert kept ≥ 0
+    @assert size(wts) == ntuple(i -> size(A, kept + i), ndims(wts))
```
cossio: `@assert size(wts) == size(A)[ndims(A)-ndims(wts)+1:end]`

src/util/util.jl
cossio: Done in 9370f6a, in both `wsum` methods — spelled `size(A)[(kept + 1):end]` using the `kept` already computed on the line above.

---

src/util/util.jl
cossio: Done in 9370f6a (both methods).

---

CHANGELOG.md
```diff
@@ -4,6 +4,38 @@ All notable changes to this project will be documented in this file. The format
...
+    as given: they are never rescaled (the per-iteration Float64
+    normalization is removed without replacement), so extreme weights (near
+    `floatmax`, or needing wider-than-`Float64` accumulation) can now
+    overflow; ordinary weights are unaffected. Internal helpers such as
+    `wmean` and `∂free_energy` remain plain weighted reductions that do not
+    validate.
+  - Unweighted training with `batchsize` larger than the number of samples now
+    clamps the batchsize (as weighted training already did) instead of
+    silently performing zero iterations.
+  - `wsum(A, wts)` is available again as the internal weighted-sum kernel
```
cossio: internal function. Don't mention in CHANGELOG. Remove other similar entries from CHANGELOG

src/util/util.jl
```diff
@@ -1,40 +1,75 @@
...
-    if dims === (:)
-        @assert size(wts) == size(A)
-        w = wts
+The reduced dimensions are inferred from the shape of `wts`, which must match
+the trailing dimensions of `A` (all of `A` for a full reduction, returning a
+scalar). Reduced dimensions are dropped from the result.
+"""
+function wsum(A::AbstractArray, wts::AbstractArray{<:Real})
+    kept = ndims(A) - ndims(wts)
+    @assert kept ≥ 0
```
cossio: remove; implied by next assert

src/util/util.jl
```diff
@@ -1,40 +1,75 @@
...
+    @assert size(wts) == size(A)[(kept + 1):end]
+    if kept == 0
```
cossio: Instead of this, have a method:

wsum(A::AbstractArray{<:Any,N}, wts::AbstractArray{<:Real,N}) where {N}

that handles the case where ndims(A) == ndims(wts)

Then the current method:

wsum(A::AbstractArray, wts::AbstractArray{<:Real})

can assume ndims(A) != ndims(wts) and needs only handle this case.

CHANGELOG.md
cossio: Done in bcd3fe6: the `wsum` bullet is gone, as is the sentence about internal helpers remaining unvalidated, and the remaining entries no longer name internal functions or the concrete lazy-weights type — they describe the visible behavior only.

---

src/util/util.jl
cossio: Done in bcd3fe6. One consequence: `wts` with more dimensions than `A` now errors slicing the size tuple (`BoundsError`) instead of `AssertionError` — the test pinning that case is updated.

---

src/util/util.jl
cossio: Done in cb4ca40, exactly as proposed: `wsum(A::AbstractArray{<:Any, N}, wts::AbstractArray{<:Real, N}) where {N}` handles the full reduction with a plain `@assert size(wts) == size(A)`, and the trailing-reduction method drops its `kept == 0` branch. One wrinkle: the `Ones` fast path has to mirror the split into two methods as well — a single `wsum(A, wts::Ones{<:Real})` would be dispatch-ambiguous against the same-`N` generic method (neither is strictly more specific). Verified with the JLArrays device test that the same-`N` `Ones` method keeps the full uniform reduction on device.

---

src/util/util.jl
```diff
@@ -1,40 +1,80 @@
...
+    S = sum(float, A; dims = (kept + 1):ndims(A))
+    return reshape(S, ntuple(d -> size(A, d), Val(kept)))
+end
+
+# Float the operands (a no-op for float arrays): narrow-integer data and
+# weights would otherwise accumulate in their own narrow type (wrapping on
+# overflow), and mixed integer/float operands would take a different matmul
+# kernel than float-converted copies, breaking exact reproducibility.
+_asfloat(A::AbstractArray{<:AbstractFloat}) = A
+_asfloat(A::AbstractArray) = float.(A)
```
cossio: these two lines can be simplified to:

```
_asfloat(A::AbstractArray) = float(A)
```

since `float(A)` does nothing if `A::AbstractArray{<:AbstractFloat}` automatically.

Do you agree? If you agree,  replace the two lines with

_asfloat(A::AbstractArray) = float(A)

src/train/initialization.jl
```diff
@@ -140,16 +169,15 @@ Initializes `rbm.w` such that typical inputs to hidden units are λ.
...
-        @assert length(wts) == size(data)[end]
-        x = reshape(data, length(rbm.visible), size(data)[end])
-        d = dot(x * Diagonal(wts), x / sum(wts))
-    end
+    @assert length(wts) == size(data)[end]
+    _validate_weights(wts)
+    # float the operands (a no-op for float arrays): narrow integer data or
+    # weights (e.g. Int8 spins) would wrap in the products otherwise
+    x = _asfloat(reshape(data, length(rbm.visible), size(data)[end]))
+    d = dot(x .* reshape(_asfloat(wts), 1, :), x / sum(wts))
```
cossio: > _asfloat(wts)

What is this buying us here ? If `wts` is `Trues`, then `_asfloat` does nothing. Otherwise, it's likely floating already. We will never have Int8 wts !! 

Find similar uses of `_asfloat` and purge them.

Do you agree? If there is something I'm missing, pushback.

src/util/util.jl
cossio: I agreed at first and tried it (589f50e) — you're right that `float(A)` is the identity on `AbstractArray{<:AbstractFloat}` — but CI caught a case where the two forms differ: `float(A)` **errors** on abstractly-typed arrays ("`float` not defined on abstractly-typed arrays"; `float(::Vector{Real})` throws), while the broadcast `float.(A)` converts them elementwise. `Real[…]` weights are accepted by the `AbstractVector{<:Real}` signatures and covered by a test, so the single-method form broke `test/util.jl`. Reverted in e0792f1, keeping the two methods with a comment explaining exactly this. If you'd rather drop support for abstract-eltype arrays here instead, the single-method form works and I can remove the test — say the word.

---

src/train/initialization.jl
```diff
@@ -140,16 +169,15 @@ Initializes `rbm.w` such that typical inputs to hidden units are λ.
...
-    else
-        @assert length(wts) == size(data)[end]
-        x = reshape(data, length(rbm.visible), size(data)[end])
-        d = dot(x * Diagonal(wts), x / sum(wts))
-    end
+    @assert length(wts) == size(data)[end]
+    _validate_weights(wts)
+    # float the operands (a no-op for float arrays): narrow integer data or
+    # weights (e.g. Int8 spins) would wrap in the products otherwise
+    x = _asfloat(reshape(data, length(rbm.visible), size(data)[end]))
```
cossio: Similar worry as below. I don't see any real scenario where this `_asfloat` would buy us anything useful. Do you ?

cossio: If `data` is Int8 (which I think was the motivation originally), it will be promoted as needed in the line below when it is multiplied to `wts`. Right ?

src/train/initialization.jl
cossio: Agreed for this site — removed in ba55cbd (verified: with the data factor floated by the division, `Int`/`Bool`/`Real[]` weights all promote correctly in the broadcast). On "find similar uses and purge": the weight-side `_asfloat` remains only on the matmul paths (`wsum`, `_weighted_outer`), where it is load-bearing — `A * Real[…]` matmul produces eltype `Any` (verified; a `Diagonal` version of this was a confirmed break), and float-converting integer weights keeps the reduction on the BLAS kernel, bit-identical to float-weight runs. Happy to purge those too if we also declare abstract-eltype weights out of scope.

---

src/train/initialization.jl
cossio: Almost right — with one precise exception. The division by `sum(wts)` floats the second `dot` factor, so the accumulation is always float, and `Int8` data promotes against float (or `Int`) weights in `x .* wts` — I verified the results are bit-identical with and without the conversion for `Bool`/`Int8`/`Int` data. The one case that genuinely wrapped was `Int8` data × `Int8` weights (`Int8(2) * Int8(100) = -56` in the broadcast) — but per your "we will never have Int8 wts", that's out of scope, so both `_asfloat`s are gone in ba55cbd and the integer-data regression test now uses ordinary float weights.

---

src/util/util.jl
```diff
@@ -1,40 +1,81 @@
...
+    @assert size(wts) == size(A)[(kept + 1):end]
+    S = sum(float, A; dims = (kept + 1):ndims(A))
+    return reshape(S, ntuple(d -> size(A, d), Val(kept)))
+end
+
+# Float the operands (a no-op for float arrays): narrow-integer data and
+# weights would otherwise accumulate in their own narrow type (wrapping on
+# overflow), and mixed integer/float operands would take a different matmul
+# kernel than float-converted copies, breaking exact reproducibility. The
+# broadcast, unlike `float(A)`, also handles abstractly-typed arrays (`Real[…]`).
```
cossio: We don't need to support abstract-typed Real[…] ...

src/util/util.jl
```diff
@@ -1,40 +1,81 @@
...
+    return transpose(_asfloat(vec(A))) * _asfloat(vec(wts))
 end
 
-_weighted_term(a, w) = iszero(w) ? zero(a) * w : a * w
+# partial reduction over the trailing dimensions of `A`
+function wsum(A::AbstractArray, wts::AbstractArray{<:Real})
+    kept = ndims(A) - ndims(wts)
+    @assert size(wts) == size(A)[(kept + 1):end]
+    # A matmul-shaped reduction, which never materializes a weighted copy of `A`
+    S = _asfloat(reshape(A, :, length(wts))) * _asfloat(vec(wts))
```
cossio: like in the previous comment on initialize_w! ... I don't see how _asfloat can be useful here in any real scenario

src/util/util.jl
cossio: Done in c2b377f: with abstract-eltype arrays declared unsupported, `_asfloat` is purged entirely — the matmul reductions take the operands as given, and the `Real[…]` weights tests are removed. Verified the integer/Bool moments testset still passes with exact equality (the matvec accumulation coincides with the float-converted run at these sizes). Also fixed in the same commit: the `wsum` docstring had been silently detached by a comment I inserted between the `@doc` block and the method during the dispatch split — that was what broke the docs `build` job (`Cannot resolve @ref for wsum`); the comment now lives inside the method body and the docstring binds again.

---

src/util/util.jl
cossio: Purged in c2b377f, together with every other remaining use — `_asfloat` no longer exists. (The one real scenario it served here was bit-exact agreement between integer-data and float-data runs through the mixed-eltype matmul kernel; the moments testset that pins this still passes without the conversion, so it goes.)

---

cossio: Correction, with CI as the pushback you asked for: the full purge broke the weighted integer-moments testset on CI (`layers.jl:617`, `-128.0` vs `-128.00000000000003`) — integer/Bool data through the mixed-eltype matmul kernel accumulates differently from BLAS gemv at the last ulp on CI's OpenBLAS (my local check coincided by hardware luck). 7f8917a restores an inline `float(...)` on the **data** operand only in the two `wsum` matmul methods — identity for float arrays, no helper, weights stay as given — which keeps the exact-equality contract of the moments testset. If you'd rather relax that testset to `≈` and drop the conversion completely, say so.

---

src/util/util.jl
```diff
@@ -1,40 +1,74 @@
...
-    # weights annihilate their entries exactly, even non-finite ones.
-    wn = w ./ (1.0 * float(maximum(wts)))
-    return mean(_weighted_term.(A, wn); dims) ./ mean(wn)
+function wsum(A::AbstractArray{<:Any, N}, wts::AbstractArray{<:Real, N}) where {N}
+    # full reduction: `wts` spans all of `A`
+    @assert size(wts) == size(A)
+    # `transpose`, not `dot`: the documented sum is `Σ Aᵢwᵢ`, without conjugation.
+    # `float` (identity on float arrays) keeps integer/Bool samples on the same
+    # matmul kernel as float-converted runs, which reduce bit-identically.
+    return transpose(float(vec(A))) * vec(wts)
```
cossio: can't we use `dot(vec(wts), float(vec(A)))` ? That way the adjoint is taken on `wts` which is Real, so no conjugation happens.

cossio: Done in 5290de5, exactly as written — the complex-`A` test (`wmean(ComplexF64[1 + im]; wts = [1.0]) == 1 + im`) still passes, confirming no conjugation.

---

src/util/util.jl
```diff
@@ -1,40 +1,75 @@
...
-    # weights annihilate their entries exactly, even non-finite ones.
-    wn = w ./ (1.0 * float(maximum(wts)))
-    return mean(_weighted_term.(A, wn); dims) ./ mean(wn)
+function wsum(A::AbstractArray{<:Any, N}, wts::AbstractArray{<:Real, N}) where {N}
+    # full reduction: `wts` spans all of `A`
+    @assert size(wts) == size(A)
+    # `dot` conjugates its first argument, a no-op for the real `wts`, so the
+    # documented sum `Σ Aᵢwᵢ` never conjugates `A`. `float` (identity on float
+    # arrays) keeps integer/Bool samples on the same kernel as float-converted
+    # runs, which reduce bit-identically.
```
cossio: simplify comment.

src/util/util.jl
```diff
@@ -1,40 +1,75 @@
...
 end
 
-_weighted_term(a, w) = iszero(w) ? zero(a) * w : a * w
+# partial reduction over the trailing dimensions of `A`
+function wsum(A::AbstractArray, wts::AbstractArray{<:Real})
+    kept = ndims(A) - ndims(wts)
+    @assert size(wts) == size(A)[(kept + 1):end]
+    # A matmul-shaped reduction, which never materializes a weighted copy of `A`.
+    # `float` (identity on float arrays) keeps integer/Bool samples on the same
+    # matmul kernel as float-converted runs, which reduce bit-identically.
```
cossio: I don't think this comment is needed

src/util/util.jl
```diff
@@ -1,40 +1,75 @@
...
+    S = float(reshape(A, :, length(wts))) * vec(wts)
+    return reshape(S, ntuple(d -> size(A, d), Val(kept)))
+end
+
+# Uniform weights reduce as a plain sum: `float` per element accumulates
+# non-float data in float without materializing a converted copy, and the
+# lazy CPU fill would take a scalar-indexing kernel in the mixed matmul
+# when `A` is a GPU array.
```
cossio: comment not needed

src/util/util.jl
```diff
@@ -1,40 +1,75 @@
...
+function wsum(A::AbstractArray, wts::Ones{<:Real})
+    kept = ndims(A) - ndims(wts)
+    @assert size(wts) == size(A)[(kept + 1):end]
+    S = sum(float, A; dims = (kept + 1):ndims(A))
+    return reshape(S, ntuple(d -> size(A, d), Val(kept)))
+end
+
+# `A * Diagonal(vec(wts)) * B'`, the weighted outer product `Σᵢ wᵢ A[:,i] B[:,i]'`.
+# Uniform `Ones` weights reduce to the plain product: `Diagonal` of a lazy CPU
+# fill takes a scalar-indexing kernel when the factors are GPU arrays.
```
cossio: comment not needed

PR #222, opened 2026-08-09
Adopt MLUtils.DataLoader for minibatch iteration

cossio:
Supersedes #211 (closed so this session can own and monitor the PR). Rebuilt on top of #220's branch, since that merges first — the base is `claude/rbm-pr-212-niye8t`, so this PR shows only the DataLoader adoption and will retarget to `master` automatically when #220 merges.

## What changed

- `infinite_minibatches` now wraps `Iterators.cycle(DataLoader(ds; batchsize, shuffle, partial = false))`. Cycling restarts the loader when an epoch is exhausted, and the loader reshuffles on each restart, reproducing the old infinite stream with a fresh permutation per epoch; `partial = false` reproduces the fixed-size batches (trailing partial batch dropped).
- The internal helpers `nobs`, `getobs`, `shuffleobs`, and `InfiniteMinibatchIterator` are removed (none were marked `public`).
- MLUtils becomes a direct dependency (compat `0.4.13`, the version whose `DataLoader` reshuffles under `Iterators.cycle`).
- `randperm` is no longer imported (its only caller was `shuffleobs`).

## Simpler than #211, thanks to #220

Building on #220 dissolves both complications the original PR carried:

- **No `nothing`-weights seam.** #220 eliminated `wts = nothing` in favor of lazy `Ones` weights everywhere, so weights are just another array: the two dispatch methods from #211 collapse into #220's single variadic signature `infinite_minibatches(ds::AbstractArray...)`, now a 4-line wrapper.
- **No empty-stream guard.** #211 guarded `batchsize > numobs` to preserve the old zero-iterations contract. #220 changed that contract deliberately — `_train!` clamps the batchsize to one full batch before building the iterator — which is exactly `DataLoader`'s native behavior, so the guard is dropped rather than ported. (Called directly with an oversized batchsize, the loader clamps with a warning; the trainers pre-clamp and never hit it.)

Net: −41 lines on top of #220; `src/train/infinite_minibatches.jl` keeps only the wrapper plus #220's `_validate_weights`.

## Trade-off to weigh (unchanged from #211)

This trades dependency-free iterator code for MLUtils' dependency chain (NNlib, StatsBase, Tables, MLCore, …), heavier than anything currently in `[deps]`. The alternative remains extracting the old iterator into a micro-package. The comparison is now at its most favorable for MLUtils, since the wrapper is ~4 lines.

## Tests

- `test/infinite_minibatches.jl` merges #220's coverage (variadic batching, lazy `Trues` slicing) with #211's behavioral tests (cycling, joint data/weight permutation, per-epoch reshuffle), plus the clamp-with-warning edge case. The testsets for the removed helpers are dropped.
- Verified locally on Julia 1.12.6: infinite_minibatches passes; train, pcd, zero_weight_training, jlarrays, ExplicitImports, and Aqua are running and I'll report on them in the thread if anything fails (CI covers them regardless).

https://claude.ai/code/session_01Xr7ZjG4LWqejQPxaiD593M

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xr7ZjG4LWqejQPxaiD593M)_


CHANGELOG.md
```diff
@@ -34,6 +34,11 @@ All notable changes to this project will be documented in this file. The format
   - `wsum(A, wts)` is available again as the internal weighted-sum kernel
```
cossio: `wsum` is an internal function, this should not be mentioned in CHANGELOG. Remove other similar entries from  CHANGELOG

cossio: Done in 5a1af63. Removed from `Unreleased`: the `wsum` bullet, the sentence about `wmean`/`∂free_energy` remaining unvalidated reductions, and the list of removed minibatch helpers (`nobs`, `getobs`, `shuffleobs`, `InfiniteMinibatchIterator`) in the MLUtils entry — that entry now records only the new direct dependency and that training behavior is unchanged. I left the released sections (6.2.0 and earlier) as historical record. Note the first two removals edit #220's entry, so they'll show as part of this PR's diff until it retargets.

---

cossio:
CI note: the docs `build` job is red here, but the failure comes from the base branch, not this PR. `cb4ca40` (on #220's branch) inserted a `# full reduction: …` comment between `@doc raw"""…"""` and `function wsum(...)` in `src/util/util.jl`, which breaks the docstring attachment (verified: `Docs.doc` finds no docstring for the `wsum` binding). `@autodocs` then omits `wsum` from the reference page and Documenter aborts on the `[`wsum`](@ref)` link inside `wmean`'s docstring. #220's own docs build should fail identically. Fix belongs in #220 (e.g. move the comment below the signature or above the `@doc` block); this PR will re-green once that lands and the base is merged in.

---

PR #229, opened 2026-08-10
Split CI test suite into 4 parallel groups

cossio:
## Motivation

CI wall time was ~23 minutes, dominated by the ~21-minute serial test run that both the `test` job and the `downgrade` job execute in full. The test files in `test/runtests.jl` are already independent modules, so they can run as parallel CI jobs.

## Changes

- `test/runtests.jl`: the test files are now organized into 4 groups, balanced using per-file wall-time measured from a recent CI run on `master` (heaviest files: `layers.jl` ~4.7 min, `jlarrays.jl` ~2.6 min, `rbm.jl` ~2.3 min, `standardized.jl` ~2.2 min). A new `RBM_TEST_GROUP` environment variable selects a single group; when unset (or `all`), the full suite runs, so a plain local `Pkg.test()` behaves exactly as before. Modules are generated from the file lists with `@eval`, replacing the hand-written module-per-file boilerplate; the same 29 files are included either way.
- `.github/workflows/ci.yml`: both the `test` and `downgrade` jobs gain a `group: ['1'..'4']` matrix dimension and pass `RBM_TEST_GROUP` to `julia-runtest`.
- `codecov.yml`: with 4 coverage uploads per commit, Codecov now waits for all of them (`after_n_builds: 4`) before computing status or commenting, so a partial merge doesn't report a spurious coverage drop.

**Observed result:** the full matrix (8 jobs) completes in ~12 minutes of wall time, versus ~23 minutes before — see [this run](https://github.com/cossio/RestrictedBoltzmannMachines.jl/actions/runs/31384513621).

## Notes

- Job names change (e.g. `Julia 1 - ubuntu-latest - x64` → `Julia 1 - ubuntu-latest - x64 - group 1`). If branch protection lists the old job names as required status checks, that list needs updating after this merges.
- The first two runs hit a transient Aqua `Persistent tasks` failure in the downgrade group-4 job: Aqua's wrapper-precompile subprocess (which inherits `--check-bounds=yes --code-coverage` and therefore recompiles every dependency from scratch) died on runners with cold per-group caches. It passed once the group's cache was warm, and the wrapper environment resolves and precompiles cleanly when reproduced locally with min-pinned deps.
- No CHANGELOG entry since this is CI/test tooling with no user-facing effect.

https://claude.ai/code/session_01YE1r2xKfPNfpoyQ3Z3JKtw

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ permissions:
   contents: read
 jobs:
   test:
-    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - group ${{ matrix.group }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -19,6 +19,12 @@ jobs:
           - ubuntu-latest
         arch:
           - x64
+        # Test groups defined in test/runtests.jl, run as parallel jobs.
+        group:
+          - '1'
+          - '2'
+          - '3'
+          - '4'
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
       - uses: julia-actions/setup-julia@fa02766e078afaaf09b14210362cee14137e6a32
@@ -30,6 +36,8 @@ jobs:
       - uses: julia-actions/julia-runtest@6e050c8013b833b1195105ff2fce9cd802f53271
         with:
           coverage: true
+        env:
+          RBM_TEST_GROUP: ${{ matrix.group }}
       - uses: julia-actions/julia-processcoverage@03114f09f119417c3242a9fb6e0b722676aedf38
       - uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f
         with:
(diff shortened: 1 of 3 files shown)


cossio:
Closing this for now: the speedup works (CI wall time ~23 min → ~12 min, all 8 matrix jobs green), but the added moving parts — hand-balanced group lists in `runtests.jl`, an 8-job matrix, and the Codecov `after_n_builds` coordination — feel like too much machinery for a 2x gain at the suite's current size.

For the record, the alternatives considered:

1. **Static hand-balanced matrix** (this PR). GitHub Actions' native mechanism — `strategy.matrix` plus an env var (`RBM_TEST_GROUP`) selecting a group of test files. This is also the established convention in the Julia ecosystem (e.g. SciML packages' `GROUP` variable). Groups were balanced from per-file wall times measured on a `master` run. Downside: the lists need occasional manual rebalancing, and coverage/status reporting needs `after_n_builds` so Codecov waits for all group uploads.
2. **Timing-based automatic splitting** (e.g. the `split-tests` or `action-split-with-timing` actions). Shards are computed from a JUnit XML timing report of previous runs, so balancing maintains itself. For Julia this would additionally require TestReports.jl to emit JUnit XML and artifact plumbing to persist timings between runs — more infrastructure than the manual lists it would replace, at 29 test files.
3. **ReTestItems.jl** — the principled Julia-native option. Tests become `@testitem` blocks and the runner schedules them dynamically across worker processes (`runtests(...; nworkers=4)`); CI stays a single job per configuration and the group lists disappear entirely. Requires migrating the test files (each testitem declares its own imports; shared helpers move to `@testsetup`). Our files are already independent modules, so the migration is mechanical, but it's a real change to the suite.

Whichever route, the ceiling is similar: duplicated per-job/per-worker setup and package loading plus the longest single file (`layers.jl` ~5 min) put a floor of roughly 8 minutes on any splitting scheme — so all options land in the 8–12 min range, i.e. about 2x. If the serial ~23 min becomes painful later, ReTestItems.jl (option 3) is the recommended path rather than reviving this branch.

One operational note from this experiment: with cold per-group caches, Aqua's `Persistent tasks` check can transiently fail in the downgrade job (its precompile subprocess inherits `--check-bounds=yes --code-coverage` and recompiles every dependency from scratch); it stabilizes once caches are warm and reproduces cleanly nowhere else.

---

PR #230, opened 2026-08-10
Make Claude PR review explicitly requested instead of automatic

cossio:
Reviews should happen when asked for, not on every push. This drops the automatic `pull_request` trigger while keeping the review itself byte-identical to what ran before.

## Changes

**`.github/workflows/claude-code-review.yml`** — same job, new trigger.

- `on: pull_request [opened, synchronize, ready_for_review, reopened]` → `on: issue_comment, pull_request_review_comment`, gated on the phrase `@review-claude`.
- The `prompt:`, `plugins:`, `plugin_marketplaces:` and permissions are unchanged. The PR number now comes from `github.event.issue.number || github.event.pull_request.number` (exactly one is set, depending on which comment event fired), so the prompt string Claude receives is the same as the one the automatic workflow built. Passing the PR explicitly also means the plugin fetches the diff from the API rather than depending on the checkout ref — which matters here, since comment events check out the default branch rather than the PR head.
- `issue_comment` fires for plain issues as well as PRs, so the `if:` also requires `github.event.issue.pull_request`; the prompt needs a PR number.

**`.github/workflows/claude.yml`** — comment only, no behavior change. The `if:` is untouched; only the note next to the plugin inputs is updated to point at `@review-claude` for a full review.

**`CLAUDE.md`** — the "PRs receive automated review comments from … the default Claude App workflow" line was no longer true, so it now says a Claude review is requested with `@review-claude`.

## Why `@review-claude` and not `@claude-review`

`contains()` in GitHub expressions is plain substring matching. `@claude-review` contains `@claude`, so that phrase would satisfy `claude.yml`'s gate too and one comment would be answered by two workflows. `@review-claude` does not contain `@claude` (the preceding character is `-`), so the two workflows stay mutually exclusive with no exclusion clause needed on either side.

## Access control

Neither workflow gates on `author_association`; both gate on the trigger phrase alone. `claude-code-action` already refuses actors without write access to the repository, and blocks GitHub Apps and bots by default, before it makes any Claude API call — so a workflow-level author check would only duplicate that, at the cost of the two Claude workflows behaving differently from each other.

The old `if:` skipped drafts and fork PRs because fork `pull_request` runs cannot reach `CLAUDE_CODE_OAUTH_TOKEN`. Comment events do not have that problem — they always run in base-repository context — so that guard is simply dropped rather than replaced.

## Testing

- Both workflow files parse under `yaml.safe_load`; verified the resulting `on:` keys, `permissions` maps, and the review job's `prompt`.
- Simulated both `if:` expressions against GitHub's `contains()` semantics over a matrix of events — PR conversation comment, PR inline review comment, plain-issue comment, no mention. Every case resolves to exactly one of *review only* / *mention only* / *nothing runs*; no input fires both workflows. Explicitly asserted `contains('@review-claude', '@claude') == false`, and that neither `if:` references `author_association`.
- `python3 .github/scripts/lint_agent_docs.py` → OK, 0 warnings.

The trigger itself can only be exercised on a live PR — commenting `@review-claude` on this one is the real check, and is also the only remaining way to get a Claude review on it.

## Notes

No `CHANGELOG.md` entry: per `CLAUDE.md`, CI and workflow plumbing is not user-facing.

https://claude.ai/code/session_01Ht8sUxuwXoSaAdZzUGAprP

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -39,8 +39,9 @@ jobs:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
 
           # Install the same plugins as claude-code-review.yml so their slash
-          # commands can be invoked by hand from a comment, e.g.
+          # commands can also be invoked ad hoc from a freeform mention, e.g.
           #   @claude /code-review:code-review --comment
+          # For the canonical whole-PR review, prefer `@review-claude`.
           plugin_marketplaces: 'https://github.com/anthropics/claude-code.git'
           plugins: 'code-review@claude-code-plugins'
 
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -39,9 +39,8 @@ jobs:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
 
           # Install the same plugins as claude-code-review.yml so their slash
-          # commands can also be invoked ad hoc from a freeform mention, e.g.
+          # commands can be invoked by hand from a comment, e.g.
           #   @claude /code-review:code-review --comment
-          # For the canonical whole-PR review, prefer `@review-claude`.
           plugin_marketplaces: 'https://github.com/anthropics/claude-code.git'
           plugins: 'code-review@claude-code-plugins'
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -94,10 +94,10 @@ CUDA.jl, and HDF5 persistence. It requires Julia 1.12 or later.
   internal refactors with unchanged observable behavior, test-only changes,
   docs, formatting, CI, workflows, agent plumbing, or other repository
   tooling.
-- PR reviews are requested explicitly, not posted on every push. Comment
-  `@review-claude` for a Claude review (runs
-  `.github/workflows/claude-code-review.yml`) or `@codex review this` for a
-  Codex Cloud review. Address each actionable finding or explain the
+- PR reviews are not automatic, and requesting one is not your call: the
+  repository owner triggers a Claude or Codex Cloud review when they want
+  one. Never trigger a review yourself, and do not ask for one. When review
+  comments do arrive, address each actionable finding or explain the
   disagreement in its thread, reply to every thread, and resolve it once
   addressed.
 - Follow `REVIEW.md`; flag substantial avoidable complexity only when a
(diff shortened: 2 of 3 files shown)


.github/workflows/claude.yml
```diff
@@ -39,8 +39,9 @@ jobs:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
 
           # Install the same plugins as claude-code-review.yml so their slash
-          # commands can be invoked by hand from a comment, e.g.
+          # commands can also be invoked ad hoc from a freeform mention, e.g.
```
cossio: why this change ? no need

cossio: Agreed, no need — reverted in eb437f0. It was only a reworded comment, no functional change, so `claude.yml` is now untouched by this PR and the diff is just `claude-code-review.yml` plus the `CLAUDE.md` line.

---

CLAUDE.md
```diff
@@ -94,10 +94,11 @@ CUDA.jl, and HDF5 persistence. It requires Julia 1.12 or later.
...
   tooling.
-- PRs receive automated review comments from Codex Cloud and the default Claude
-  App workflow in `.github/workflows/claude-code-review.yml`. Address each
-  actionable finding or explain the disagreement in its thread, reply to every
-  thread, and resolve it once addressed.
+- PRs receive automated review comments from Codex Cloud. A Claude review is
+  requested explicitly, not on every push: comment `@review-claude` on the PR
+  to run `.github/workflows/claude-code-review.yml`. Address each actionable
+  finding or explain the disagreement in its thread, reply to every thread,
+  and resolve it once addressed.
```
cossio: Claude should not trigger any review. I'll trigger them myself if I want to. But Claude should be prepared to address the review comments.

cossio: Fixed in 1200acf. My wording had documented the trigger phrases to agents, which invited exactly what you're ruling out. Now:

> PR reviews are not automatic, and requesting one is not your call: the repository owner triggers a Claude or Codex Cloud review when they want one. Never trigger a review yourself, and do not ask for one. When review comments do arrive, address each actionable finding or explain the disagreement in its thread, reply to every thread, and resolve it once addressed.

I also dropped `@review-claude` and `@codex review this` from `CLAUDE.md` entirely, so the phrases aren't sitting in an agent's context as an available action. `@review-claude` stays documented in the header comment of `.github/workflows/claude-code-review.yml`, where it describes the workflow rather than instructing an agent.

---

PR #235, opened 2026-08-28
Add TestCompatHygiene checks to the test suite

cossio:
Adds [TestCompatHygiene.jl](https://github.com/cossio/TestCompatHygiene.jl) to the test suite, as requested.

## Changes

- `test/Project.toml`: add `TestCompatHygiene` to `[deps]` with `[compat] TestCompatHygiene = "1"`.
- `test/compat_hygiene.jl`: new test file calling `TestCompatHygiene.test_all(RestrictedBoltzmannMachines)`, mirroring the Aqua/ExplicitImports style of the existing hygiene tests.
- `test/runtests.jl`: include the new file as an independent `compat_hygiene_tests` module, next to `aqua_tests`.

The check guards the workspace invariant that `test/Project.toml` must not declare `[compat]` bounds for names the root `Project.toml` already owns (deps, compat entries including `julia`, or the package itself), since workspace members share one manifest and repeated bounds are intersected, silently narrowing the root's declared resolve.

No CHANGELOG entry: test-only change.

## Testing

- `julia --project=test test/compat_hygiene.jl` passes standalone (1/1): the current `test/Project.toml` `[compat]` entries are all for test-only dependencies, none owned by the root project.

https://claude.ai/code/session_01PpiKWyR18iV7w3qhsRLBgz

---
_Generated by [Claude Code](https://claude.ai/code/session_01PpiKWyR18iV7w3qhsRLBgz)_


cossio:
**CI status: the "Downgrade compat - Julia min" failure is not caused by this PR.** It failed twice (original run + one re-run) in Aqua's *Persistent tasks* check: the precompile subprocess exits silently without writing its `done.log` sentinel (`Aqua persistent_tasks.jl:118`). Everything else in the job passes, including the new TestCompatHygiene check.

Evidence that it's decoupled from this diff:

- The diff is test-only. Aqua's persistent-tasks check builds a throwaway wrapper env from the **root** `Project.toml` (untouched here) and re-resolves it fresh from the registry at test time — `test/Project.toml` plays no role in it.
- I diffed the resolved test sandbox of this PR's failing job against master's passing run from this morning (07:04 UTC, commit `1f994ad` — this PR's exact base): **all 177 package versions are identical**; the only delta is the addition of `TestCompatHygiene v1.0.0` itself (deps: TOML and Test stdlibs only, loaded after the aqua testset). Same Julia 1.12.7, same Aqua v0.8.14.
- I could not reproduce locally on Julia 1.12.7 with Aqua 0.8.14 or 0.8.16, against both this branch and master, with pristine or minimum-pinned root compat — `test_persistent_tasks` passes in every combination.

Since the wrapper env floats with the live registry while the sandbox is pinned by the downgrade action's resolver, the failure window (passing 07:04 UTC, failing deterministically from 20:55 UTC on identical inputs) points at registry/runner drift during the day — most likely a transitive release that makes the wrapper's fresh precompile die silently (the subprocess prints nothing, consistent with an OOM kill rather than a Julia error). I'd expect the same failure on master's next CI run. No upstream fix exists yet to port (Aqua 0.8.16's persistent-tasks code is identical to 0.8.14's).

I've spent the one confirming re-run. The PR is otherwise green (coverage unchanged, all other jobs passing) with auto-merge armed; I'll keep watching it. A later re-run of the failed job may go green on its own once the drift resolves.

---

PR #238, opened 2026-09-20
Simplify shared code paths across layers, RBM variants, and trainers

cossio:
## Summary

Whole-codebase readability pass over `src/` and `ext/` (a `/simplify` review: reuse, simplification, efficiency, altitude). Net −100 lines. No intended behavior change, with one small user-visible exception noted below.

### Duplicated logic collapsed
- **`pcd!`**: the argument validation and the per-iteration gradient step were copied verbatim across the `RBM`, `CenteredRBM`, and `StandardizedRBM` trainers. They now live in `_pcd_check_args` and `_pcd_step!` (`src/train/pcd.jl`); each `pcd!` keeps only its model-specific centering/standardization and gauge lines.
- **dReLU reparameterizations**: pReLU, xReLU, and nsReLU each carried eight one-line forwarders to the dReLU statistics (nsReLU forwarding through xReLU). One block on the `_dReLUReparam` union in `common.jl` replaces the three. The pReLU/xReLU conversions from dReLU share `_drelu_shared_params`, and their `∂energy_from_moments` share `_prelu_∂θγΔ` (same expressions, same rounding).
- **Weight regularization**: `∂regularize!` for `RBM` and for the offset models computed the same l1/l2/l2l1 terms; `_∂regularize_weights!` now does it once, with an optional weight scale. The per-layer `∂regularize_add_visible_offset!` dispatch is replaced by a row sum over the field-regularization array, which is exact because only the field rows are nonzero. The offset version also skips building the equivalent plain RBM when no regularization is requested.
- **Copies removed**: `log_partition(::StandardizedRBM)`, `weight_norms(::StandardizedRBM)`, `potts_to_gumbel`/`gumbel_to_potts` for `StandardizedRBM`, `rescale_hidden_activations!`, `moments_from_inputs` for `Gaussian`/`ReLU`, the three HDF5 `_load_rbm` bodies, `∂cgf` (now `batchmean_moments`), `categorical_sample_from_logits_gumbel` (now `onehot_decode`), `tnvar`, and the hand-rolled `batchvar`/`total_mean_h_from_v`/`zeros_like`/`ones_like` re-implementations.

### Readability
- Named the recurring layer unions: `_PottsLayers`, `_ThetaLayers`, `_dReLUFamily`, `_dReLUReparam`.
- `_along_visible` / `_along_hidden` for the recurring `reshape(x, map(one, size(...))..., size(...)...)` broadcasts against `rbm.w` (eight sites).
- `_sum_layer_dims` for the "sum over layer dims, keep batch dims" reduction shared by `energy` and `cgf`.
- `interaction_energy` had two branches with identical bodies; `∂interaction_energy` had three. Both are now two-way. `batch_size(rbm, v, h)` no longer special-cases empty batch tuples since `join_batch_size` handles them.
- Potts ↔ PottsGumbel layer conversions use dispatch instead of runtime `isa`.
- `CenteredRBM`'s constructor mirrors the `StandardizedRBM` one instead of spelling out its type parameters twice.

### User-visible change
- `log_partition` (exhaustive enumeration) and `log_likelihood` are now untyped on the RBM argument, so they also accept a `CenteredRBM`. The Gaussian–Gaussian analytic `log_partition` method is unchanged. Recorded under `## Unreleased` in `CHANGELOG.md`.

### Test-only fix for a failure that is pre-existing on `master`
- `test/train.jl`: CI now resolves Julia `'1'` to 1.13.0, where the seeded `standardized pcd potts moment matching` case ends at moment gaps of 0.054 against the 0.05 threshold (the trained parameters are bit-identical on `master`, and `master`'s own CI re-run 33555505865 fails the same way). With 5 Gibbs steps per update the final-iterate gaps range from 0.02 to 0.055 across seeds; 20 steps bring the worst gap over seeds 60–75 to 0.032 (8973332).

### Known red check, handled separately
- The `Downgrade compat - Julia min` job fails only on Aqua's `Persistent tasks` check. That job resolves Aqua at the test compat floor, 0.8.14, whose check re-resolves its wrapper environment and, on a cold depot cache, misreports a silent wrapper-compile failure as a persistent task (Aqua issue #315, fixed upstream in 0.8.17 by PR #389). It fails identically on `master`'s re-run. The fix is a compat-floor bump to Aqua 0.8.18 in `test/Project.toml`, which will go in a separate PR; details in the comments below.

### Skipped findings (deliberately not applied)
- A `_field_rows` trait to replace the per-layer `shift_fields`/`anneal_zero` methods: `shift_fields` is differentiated with Zygote through `uncenter`/`regularization_penalty` in the tests, so it must stay non-mutating; the trait form would not be.
- Unifying `center_*` with `standardize_*`, the `initialize!` defaults via `@declare_layer`, `drelu_mode`, the HDF5 name tables, and the larger performance items (in-place `anneal!`, fused dReLU moment kernels, `metropolis_once` free-energy reuse): these either touch numerics, the compatibility-sensitive file format, or exceed a readability pass.

## Test plan
- [x] Locally on Julia 1.13: `rbm`, `layers`, `regularize`, `shift_fields`, `onehot`, `truncnorm`, `partition`, `centered`, `standardized`, `pcd`, `hdf5`, `pottsgumbel`, `jlarrays`, `explicit_imports`, `aqua`, `ais`, `initialization`, `gauge/zerosum`, `gauge/rescale_hidden`, and `train` all pass.
- [x] CI, Julia 1.13 job: green on 5552478 and 5fc34c2 (Codecov patch and project green).
- [x] CI, downgrade job on 5fc34c2: every package test passes on Julia 1.12.7; only Aqua's persistent-tasks check fails (see above).
- [x] Runic 1.7 `--check` passes on `src/` and `ext/`.

https://claude.ai/code/session_013oikaG8MAfPuoBRyAdQqKX


src/layers/nsReLU.jl
```diff
@@ -19,14 +19,8 @@ corresponding unit. This is the `xReLU` potential with ``\gamma = 1``.
...
-energies(layer::nsReLU, x::AbstractArray) = energies(xReLU(layer), x)
-cgfs(layer::nsReLU, inputs::AbstractArray = Falses(size(layer))) = cgfs(xReLU(layer), inputs)
-sample_from_inputs(layer::nsReLU, inputs::AbstractArray = Falses(size(layer))) = sample_from_inputs(xReLU(layer), inputs)
-mode_from_inputs(layer::nsReLU, inputs::AbstractArray = Falses(size(layer))) = mode_from_inputs(xReLU(layer), inputs)
-mean_from_inputs(layer::nsReLU, inputs::AbstractArray = Falses(size(layer))) = mean_from_inputs(xReLU(layer), inputs)
-var_from_inputs(layer::nsReLU, inputs::AbstractArray = Falses(size(layer))) = var_from_inputs(xReLU(layer), inputs)
-meanvar_from_inputs(layer::nsReLU, inputs::AbstractArray = Falses(size(layer))) = meanvar_from_inputs(xReLU(layer), inputs)
-mean_abs_from_inputs(layer::nsReLU, inputs::AbstractArray = Falses(size(layer))) = mean_abs_from_inputs(xReLU(layer), inputs)
+# The statistics (`energies`, `cgfs`, sampling, ...) are those of the equivalent dReLU
+# layer; see common.jl.
```
cossio: this comment is not needed

src/layers/pottsgumbel.jl
```diff
@@ -12,18 +12,4 @@ function sample_from_inputs(layer::PottsGumbel, inputs::AbstractArray = Falses(s
...
-end
-
-function gumbel_to_potts(layer::AbstractLayer)
-    if layer isa PottsGumbel
-        return Potts(layer)
-    else
-        return layer
-    end
-end
+# `potts_to_gumbel` / `gumbel_to_potts` are defined in potts.jl, once both types exist.
```
cossio: comment not needed

src/layers/prelu.jl
```diff
@@ -37,14 +37,16 @@ function _validate_layer_parameters(layer::pReLU)
...
-energies(layer::pReLU, x::AbstractArray) = energies(dReLU(layer), x)
-cgfs(layer::pReLU, inputs::AbstractArray = Falses(size(layer))) = cgfs(dReLU(layer), inputs)
-sample_from_inputs(layer::pReLU, inputs::AbstractArray = Falses(size(layer))) = sample_from_inputs(dReLU(layer), inputs)
-mean_from_inputs(layer::pReLU, inputs::AbstractArray = Falses(size(layer))) = mean_from_inputs(dReLU(layer), inputs)
-var_from_inputs(layer::pReLU, inputs::AbstractArray = Falses(size(layer))) = var_from_inputs(dReLU(layer), inputs)
-meanvar_from_inputs(layer::pReLU, inputs::AbstractArray = Falses(size(layer))) = meanvar_from_inputs(dReLU(layer), inputs)
-mode_from_inputs(layer::pReLU, inputs::AbstractArray = Falses(size(layer))) = mode_from_inputs(dReLU(layer), inputs)
-mean_abs_from_inputs(layer::pReLU, inputs::AbstractArray = Falses(size(layer))) = mean_abs_from_inputs(dReLU(layer), inputs)
+# The statistics (`energies`, `cgfs`, sampling, ...) are those of the equivalent dReLU
+# layer; see common.jl.
```
cossio: comment not needed

src/layers/xrelu.jl
```diff
@@ -19,28 +19,20 @@ the corresponding unit. This is the `pReLU` potential with the bounded asymmetry
...
-energies(layer::xReLU, x::AbstractArray) = energies(dReLU(layer), x)
-cgfs(layer::xReLU, inputs::AbstractArray = Falses(size(layer))) = cgfs(dReLU(layer), inputs)
-sample_from_inputs(layer::xReLU, inputs::AbstractArray = Falses(size(layer))) = sample_from_inputs(dReLU(layer), inputs)
-mode_from_inputs(layer::xReLU, inputs::AbstractArray = Falses(size(layer))) = mode_from_inputs(dReLU(layer), inputs)
-mean_from_inputs(layer::xReLU, inputs::AbstractArray = Falses(size(layer))) = mean_from_inputs(dReLU(layer), inputs)
-var_from_inputs(layer::xReLU, inputs::AbstractArray = Falses(size(layer))) = var_from_inputs(dReLU(layer), inputs)
-meanvar_from_inputs(layer::xReLU, inputs::AbstractArray = Falses(size(layer))) = meanvar_from_inputs(dReLU(layer), inputs)
-mean_abs_from_inputs(layer::xReLU, inputs::AbstractArray = Falses(size(layer))) = mean_abs_from_inputs(dReLU(layer), inputs)
+# The statistics (`energies`, `cgfs`, sampling, ...) are those of the equivalent dReLU
+# layer; see common.jl.
```
cossio: comment not needed

src/layers/nsReLU.jl
cossio: Removed in 313c196.

---

src/layers/pottsgumbel.jl
cossio: Removed in 313c196.

---

src/layers/prelu.jl
cossio: Removed in 313c196.

---

src/layers/xrelu.jl
cossio: Removed in 313c196.

---

src/train/gradient.jl
```diff
@@ -58,23 +58,18 @@ function ∂interaction_energy(
...
-        hflat = with_eltype_of(rbm.w, vec(batchmean(rbm.hidden, h; wts)))
-        ∂wflat = -vflat * hflat'
-    elseif ndims(rbm.hidden) == ndims(h)
-        vflat = with_eltype_of(rbm.w, vec(batchmean(rbm.visible, v; wts)))
-        hflat = with_eltype_of(rbm.w, vec(h))
-        ∂wflat = -vflat * hflat'
-    else
-        # weighted batch average as a Diagonal-weighted matmul, as in `batchcov`
+    if ndims(rbm.visible) < ndims(v) && ndims(rbm.hidden) < ndims(h)
+        # both batched: weighted batch average as a Diagonal-weighted matmul, as in `batchcov`
```
cossio: comment not needed

src/train/gradient.jl
```diff
@@ -58,23 +58,18 @@ function ∂interaction_energy(
...
         vflat = with_eltype_of(rbm.w, flatten(rbm.visible, v))
         hflat = with_eltype_of(rbm.w, flatten(rbm.hidden, h))
         ∂wflat = -_weighted_outer(vflat, wts, hflat) / sum(wts)
+    else
+        # at most one side is batched: average it, then take the outer product
```
cossio: comment not needed

src/train/pcd.jl
```diff
@@ -3,6 +3,49 @@ function _default_fantasy_chains(rbm, batchsize::Int)
...
+    length(wts) == size(data, ndims(data)) ||
+        throw(DimensionMismatch("length(wts) must equal the number of data samples"))
+    validate_wts(wts)
+    return mean(wts), min(batchsize, length(wts))
+end
+
+# One PCD parameter update: positive phase on the minibatch `(vd, wd)`, negative phase on
+# the persistent chains `vm` (updated in place), regularization, and optimiser step.
+# Returns the updated optimiser state and parameters, and the gradient fed to the
+# optimiser. `regularization` keywords are forwarded to `∂regularize!`.
```
cossio: comment not needed

cossio:
**CI failure on the first head (`Julia 1 - ubuntu-latest`)**: the only failing case was the seeded `standardized pcd potts moment matching` testset in `test/train.jl` (gaps 0.0545 / 0.0527 / 0.0530 against the 0.05 threshold). It is not introduced by this PR: under the same seed the trained `w`, `visible.par`, and `hidden.par` hash identically on `master` and on this branch. What changed is the toolchain: master's last green run used Julia 1.12.7, while CI now resolves `'1'` to 1.13.0, whose RNG stream puts this seed at 0.054 on `master` as well.

With 5 Gibbs steps per update the final-iterate gaps range from 0.02 to 0.055 across nearby seeds, so the threshold sits inside the test's noise floor. Commit 8973332 raises the steps for this case to 20, which brings the worst gap over seeds 60–75 to 0.032 (seed 63: 0.008 / 0.007 / 0.010). The other train.jl cases are unchanged.

---

src/train/gradient.jl
cossio: Removed in 5552478.

---

src/train/gradient.jl
cossio: Removed in 5552478.

---

src/train/pcd.jl
cossio: Removed in 5552478.

---

src/regularize.jl
```diff
@@ -15,23 +15,32 @@ function ∂regularize!(
     if !iszero(l2_fields)
         ∂regularize_fields!(∂.visible, rbm.visible; l2_fields)
     end
+    _∂regularize_weights!(∂.w, rbm; l1_weights, l2_weights, l2l1_weights)
+    zerosum && zerosum!(∂, rbm)
+    return ∂
+end
+
+# Adds the weight-regularization gradient of `rbm` to `∂w`, divided by `scale` (the
+# weight scales of an equivalent standardized model; lazy ones for a plain `RBM`).
```
cossio: comment not needed

cossio: Removed in 95cfc5f.

---

src/standardized.jl
```diff
@@ -83,12 +83,8 @@ Absorbs `scale_h` into the hidden layer if it has a scale parameter, returning `
 if this was done. The modified RBM is equivalent to the original one.
 """
 function rescale_hidden_activations!(rbm::StandardizedRBM)
-    if rescale_activations!(rbm.hidden, rbm.scale_h)
-        rbm.offset_h ./= rbm.scale_h
-        rbm.scale_h ./= rbm.scale_h
-        return true
-    end
-    return false
+    # copy: `rescale_hidden!` divides `scale_h` itself by the scaling factor
```
cossio: comment not needed

cossio: Removed in 5fc34c2.

---

cossio:
**CI failure on 8973332 (`Downgrade compat - Julia min`, Julia 1.12.7)**: every package test passed, including the adjusted `standardized pcd potts moment matching` case (5/5 on 1.12). The one red item is Aqua's `Persistent tasks` check, which reported `done.log was not created, but precompilation exited`: its precompile subprocess died before writing its marker, so this is the subprocess failing rather than a task left running. The same source passed that check in the Julia 1.12 job on 4934967 a few minutes earlier and in the Julia 1.13 job on 5552478; the commits in between only remove comments and change the test's `steps` count, nothing that touches precompilation. The downgrade job is re-running on identical source for 5552478 and 5fc34c2; if it reproduces there I'll dig into it.

---

cossio:
**Root cause of the `Downgrade compat` failures, and fix pushed (34815d6).** The check reproduced on 5552478, so I dug in and reproduced it locally on Julia 1.12.7.

The downgrade job resolves Aqua at the compat floor, **0.8.14**, whose `persistent_tasks` check builds its wrapper environment with `Pkg.develop` after `Pkg.respect_sysimage_versions(false)`. When the restored depot cache is cold for that environment, the wrapper's precompile subprocess compiles registry versions of stdlibs (Dates, TOML, UUIDs, ...) that differ from the ones already loaded in that session; the wrapper package itself then fails to compile with `Module ChainRulesCore ... is missing from the cache`, which Pkg reports only as "1 dependencies failed but may be precompilable after restarting julia" and **exits 0**. No done marker is written and Aqua reports `done.log was not created, but precompilation exited`. With a warm cache (or on a second pass) the same wrapper compiles and the check passes, so the outcome depends on the CI cache state, not the source: locally the same wrapper gives `exit=0, done.log=false` cold and `done.log=true` warm.

The Julia 1 job resolves Aqua **0.8.18**, which mirrors the current manifest into the wrapper instead of re-resolving, and passes the check in ~11 s (vs ~50 s for the re-resolving wrapper). 34815d6 raises the test compat floor to `Aqua = "0.8.18"` so the downgrade job uses the same implementation. I also re-ran master's last CI run (33555505865) to check the base under today's cache; it is still in progress.

---

PR #239, opened 2026-09-20
Bump Aqua compat in the test project to 0.8.18

cossio:
## Summary

- Raise the `Aqua` compat lower bound in `test/Project.toml` from `0.8.14` to `0.8.18` (the current release in General).

## Why

The downgrade-compat CI job resolves Aqua at the compat floor. Aqua 0.8.14's persistent-tasks check builds its wrapper environment by re-resolving, so with a cold depot cache the wrapper subprocess precompiles stdlib versions that differ from the loaded ones, fails to compile, and the check reports "done.log was not created, but precompilation exited". With a warm cache it passes, which is why it flipped between runs of identical source. Aqua 0.8.18 mirrors the current manifest into the wrapper instead, so the subprocess runs against already-compiled dependencies.

This is the separate PR promised by the revert of that bump in #238.

## Verification

- `julia --project=test` instantiates and resolves `Aqua v0.8.18`.
- `test/aqua.jl` run standalone: 11/11 pass.
- After merging `master` (which carries the `steps = 20` fix for the `standardized pcd potts moment matching` test), `test/train.jl` passes standalone on Julia 1.13.

No `CHANGELOG.md` entry, since this is a test-only change.

https://claude.ai/code/session_012biyZTvpyCRjqbAQwhRpbL


test/Project.toml
```diff
@@ -29,8 +29,11 @@ TestCompatHygiene = "b0582c03-59af-4651-a67d-b5bc031813b7"
 ValueHistories = "98cad3c8-aec3-5f06-8e41-884608649ab7"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
+[sources]
+RestrictedBoltzmannMachines = {path = ".."}
+
```
cossio: why did you add this ?? This is not needed !

cossio: You're right, it was not intended. `Pkg.develop(PackageSpec(path=pwd()))` wrote that `[sources]` block into `test/Project.toml` while I was setting up the local test environment, and I committed it by mistake. Removed in 3f1a6ee; the PR diff against master is now only the `Aqua = "0.8.18"` line.

---

cossio: Correction: the commit removing the block is f2d89c7, not 3f1a6ee.

---

